PR #1, opened 2023-01-23
Create scaffolding for Garden backend

WillEngler:
This PR contains the Terraform needed to set up the Garden backend API via AWS API Gateway and Lambda functions. It also contains the body of two Python Lambdas:
- A Globus Auth authorizer function which gates access to the backend
- A skeleton backend Lambda where we can write our application logic. (Stealing the idea from some Terraform tutorials on this topic, I recommend we have just one "app" lambda that does some internal routing over standing up new Lambdas per route.)

This draft PR got to the point where I can run a clean terraform apply and get all the needed resources created. (At least in theory. Setting this down before I could test and find out how it's broken.)


WillEngler:
Breaking protocol and merging without review to get this on main and do trial and error on the GH action

PR #4, opened 2023-01-30
replaces the DataCite logic implemented in the SDK

OwenPriceSkelly:
this PR is a friend of https://github.com/Garden-AI/garden/pull/34

### DIFF
--- a/app/doi.py
+++ b/app/doi.py
@@ -1,0 +1,57 @@
+import os
+import json
+
+import requests
+from requests.exceptions import HTTPError
+
+import logging
+
+logger = logging.getLogger()
+
+
+def mint_doi(event, _context, _kwargs):
+    payload = event["body"]
+    try:
+        DATACITE_REPOSITORY_ID = os.environ["DATACITE_REPOSITORY_ID"]
+        DATACITE_PASSWORD = os.environ["DATACITE_PASSWORD"]
+        DATACITE_ENDPOINT = os.environ["DATACITE_ENDPOINT"]
+        DATACITE_PREFIX = os.environ["DATACITE_PREFIX"]
+    except KeyError as e:
+        message = (
+            "Garden server was unable to authenticate with DataCite. Please "
+            "contact support and/or open an issue at "
+            "https://github.com/Garden-AI/garden-backend/issues. "
+        )
+        logger.error(f"DATACITE_* environment variables not set. env: {os.environ}")
+        return {
+            "statusCode": 500,
+            "body": json.dumps({"message": message, "error": str(e)}),
+        }
+    try:
+        payload["data"]["attributes"]["prefix"] = DATACITE_PREFIX
+
+        res: requests.Response = requests.post(
+            DATACITE_ENDPOINT,
+            headers={"Content-Type": "application/vnd.api+json"},
+            json=payload,
+            auth=(DATACITE_REPOSITORY_ID, DATACITE_PASSWORD),
+        )
+        res.raise_for_status()
+    except KeyError as e:
+        # failed to set prefix due to malformed payload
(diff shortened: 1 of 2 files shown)


app/lambda_function.py
```diff
@@ -20,13 +25,52 @@ def hello(event, context, kwargs):
...
+                {
+                    "message": message,
+                    "error": str(e)
+                }
+            ),
+        }
+    except HTTPError as e:
+        # propagate errors from requests.raise_for_status directly
+        # DataCite error responses seem minimal as far as they're documented
+        return {"statusCode": e.response.status_code, "body": str(e)}
```
OwenPriceSkelly, line 63: this was a nit I believe

OwenPriceSkelly: should be 500(?)

app/doi.py
```diff
@@ -0,0 +1,57 @@
...
+        DATACITE_PREFIX = os.environ["DATACITE_PREFIX"]
+    except KeyError as e:
+        message = (
+            "Garden server was unable to authenticate with DataCite. Please "
+            "contact support and/or open an issue at "
+            "https://github.com/Garden-AI/garden-backend/issues. "
+        )
+        logger.error(f"DATACITE_* environment variables not set. env: {os.environ}")
+        return {
+            "statusCode": 503,
```
WillEngler, line 27: I think 503 usually means "we're temporarily down for maintenance." imo this is is a 500 because something about our service is messed up.

app/doi.py
```diff
@@ -0,0 +1,57 @@
...
+import requests
+from requests.exceptions import HTTPError
+
+import logging
+
+logger = logging.getLogger()
+
+
+def mint_doi(event, _context, _kwargs):
+    payload = event["body"]
```
WillEngler, line 13: We can get away with this in prototype mode, but this is a point where we will need to validate input coming in from outside of our system. (The call will probably be generated with our sdk, but it might not be.)

app/doi.py
OwenPriceSkelly: good catch, thanks

app/doi.py
OwenPriceSkelly: we're going to add the garden sdk as a dependency at some point, right? Once the other doi PR is up, validating could just be like: 
``` python
# from garden_ai.datacite import DataciteSchema
DataciteSchema.parse_obj(payload)
```

app/lambda_function.py
```diff
@@ -20,13 +22,4 @@ def hello(event, context, kwargs):
...
-    return {
-        'statusCode': 201,
-        'body': json.dumps(
-            {
-                "success": True,
-                'message': 'I made a DOI for you. (Not really though, sorry.)',
-            })
-    }
\ No newline at end of file
+app.route("/doi", methods=["POST"])(mint_doi)  # equivalent to decorator syntax
```
WillEngler, line 25: Thanks for setting up this pattern

WillEngler (approved):
I would switch the 503 to 500 but no blockers on my side. Nice work :)

app/doi.py
WillEngler: Interesting! I suppose so. I definitely like the idea of having a library do the validation for us. Especially if it can spit out a nice error message. ("Expected a string here, got a ...")

app/lambda_function.py
OwenPriceSkelly: np 👍

app/doi.py
OwenPriceSkelly: yeah that's a big selling point for pydantic, they've even got a CLI to automatically generate the appropriate pydantic classes for a target jsonschema (which is what I did w/ for datacite)

PR #8, opened 2023-02-26
Deploy ML Flow behind Garden's API Gateway

WillEngler:
## Summary

This PR merges what had been two distinct terraform stacks. In `infra` there was our Garden API Gateway and Lambda app stack. In `infra/oauth2-mlflow-aws` there was a stack for an instance of mlflow deployed behind oauth2-proxy. This PR removes oauth2-proxy and makes mlflow a resource under our API Gateway.

## Details

I used [this reference Terraform](https://github.com/turnerlabs/terraform-ecs-fargate-apigateway) as a jumping off point for how to deploy a Fargate app behind API Gateway.

## Shortcomings and future work

The VPC that MLFlow is deployed into is not described by the Terraform in this stack. I set up the VPC with a mix of terraform from [the old template I was working from](https://getindata.com/blog/deploying-secure-mlfow-aws/) plus manual creation of a NAT gateway and some route tables in the console. I think getting everything but the VPC defined in Terraform is good enough for now. If we have time later though, it would be nice to have _everything_ in Terraform.

Furthermore, note that the version of ML Flow running on the backend is 1.22 (also obtained from [the original tutorial](https://getindata.com/blog/deploying-secure-mlfow-aws/) I used). In late 2022 ML Flow started putting out 2.x releases. Testing with a 2.x version of the ML Flow SDK, it seems like the currently deployed backend version works ok for uploading and pulling down models - our main needs for V0. But as soon as we find a feature we need that the 1.x backend doesn't support we should upgrade. (I made a quick naive attempt to use a more recent release but it seems like some command line options may have changed.) [Here is a follow-up issue to track the upgrade.](https://github.com/Garden-AI/garden-backend/issues/7)


OwenPriceSkelly (approved):
(for what it's worth) looks good to me!

Issue #7, opened 2023-02-26
Upgrade ML Flow backend to a 2.x version


WillEngler:
We currently have ML Flow 1.22 on the backend, but as of late 2022, 2.x releases are available.

WillEngler:
The error that was thwarting me when trying an [official ML Flow image](https://github.com/mlflow/mlflow/pkgs/container/mlflow/versions?filters%5Bversion_type%5D=tagged) was that at task startup time, the task could not find the `pymysql` library. This is the driver referenced in our SQLAlchemy connection string (mysql+pymysql://${aws_rds_cluster.backend_store.master_username}). It seems like the getindata version bundles this and the official image doesn't. Maybe getting us to 2.x is as straightforward as building a wrapper container around the official image with the DB Driver we need installed.

BenGalewsky:
Try this one https://hub.docker.com/r/evk02/mlflow

WillEngler:
(building off our conversation in Gathertown): it looks like that one bundles the postgres driver in the outermost layer to make it work. ([requirements link](https://github.com/zoltan-fedor/helm-mlflow/blob/master/requirements.txt)) That helps confirm our idea of what's going on here. 

So to upgrade to 2.x we can either:
1. Use some existing community container (like the exk02 one you linked) that installs the db wrapper and switch to a matching db (postgres in this case)
2. make our own thin wrapper container

When I pick this up I'll start with (1)

PR #9, opened 2023-03-07
Allow artifact proxying in ML Flow

WillEngler:
Description:

Bumps MLFlow backend from 1.22 to 1.24. That's the latest labeled version I saw from getindata. Crucially, the ability to proxy model uploads was added in 1.23. (So users don't need permissions to hit our artifact bucket.) Also adds some permissions to the app role needed to write to S3. I had confused task roles and task execution roles in the earlier attempt.

While I was here, I gave upgrading to 2.x a more earnest try with the official MLFlow image. I didn't get that working but I'll record some lessons learned in #7

### DIFF
--- a/infra/role.tf
+++ b/infra/role.tf
@@ -13,12 +13,27 @@ resource "aws_iam_role" "app_role" {
       },
     ]
   })
+  inline_policy {
+    name = "allow_s3_access"
+
+    policy = jsonencode({
+      Version = "2012-10-17"
+      Statement = [
+        {
+          Effect   = "Allow"
+          Action   = ["s3:*Object"]
+          Resource = ["arn:aws:s3:::${aws_s3_bucket.artifacts.bucket}/*"]
+        },
+        {
+          Effect   = "Allow"
+          Action   = ["s3:ListBucket"]
+          Resource = ["arn:aws:s3:::${aws_s3_bucket.artifacts.bucket}"]
+        },
+      ]
+    })
+  }
 }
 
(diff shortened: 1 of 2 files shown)


BenGalewsky:
Hmm - I've been using MLFlow 2.0
There is an official MLFlow docker image repository at https://github.com/mlflow/mlflow/pkgs/container/mlflow

infra/role.tf
```diff
@@ -13,13 +13,28 @@ resource "aws_iam_role" "app_role" {
...
     ]
   })
-}
+  inline_policy {
+    name = "allow_s3_access"
 
-resource "aws_iam_role_policy_attachment" "ecs_execution" {
-  policy_arn = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
-  role       = aws_iam_role.app_role.name
+    policy = jsonencode({
```
OwenPriceSkelly, line 19: (just for my own understanding) what's the upshot of pulling this out of the `resource "aws_iam_role_policy" "s3"{ ...` block into its own thing?

OwenPriceSkelly (approved):

infra/role.tf
WillEngler: No real rhyme or reason on which syntax I'm going with other than the example documentation I happen to be looking at and got working at the time of writing. The motivation to change this was that I had the policy on the wrong role. (It was set up so the task _execution_ role was able to manipulate S3 objects but we actually needed the _task_ itself to talk to S3.) And then ... this is just the particular way I got it working.

PR #10, opened 2023-03-08
make PR template

WillEngler:
wow look at this unstructured PR body text. What if we had guidance from above on what content and structure it had?


WillEngler:
Going to self-merge to see if it works and then have genuine open discussion in the more trafficed `garden` repo.

Issue #11, opened 2023-03-09
Fix API Gateway <-> MLFlow integration such that that models don't get mangled in transit


WillEngler:
## Short description

Proxy models through MLFlow without corrupting them.

## Context

Switching to having MLFlow proxy model upload and download resulted in models being corrupted. It turns out it was an issue described [here](https://twitter.com/iamakulov/status/946698417807511552). API Gateway was trying to coerce the binary model payload into text, corrupting the model file in the process.

# Assumptions:
1.

# Acceptance Criteria
Given a test scenario like [this one](https://github.com/Garden-AI/garden/pull/77) when a user uploads a model and then downloads it, it will be loadable from Pickle and not full of "�" characters.

WillEngler:
Update: per AWS support and some additional docs, the HTTP flavor of API Gateway has the same coercion behavior. So might close this ticket or revise to describe the right fix to try once I land on it.

PR #12, opened 2023-03-13
Allowlist binary content-type header

WillEngler:
Addresses #11. Sibling PR with https://github.com/Garden-AI/garden/pull/80

## Overview

Tell our API Gateway that it can let requests with the Content-Type application/octet-stream pass through without getting coerced to UTF-8.

## Discussion

Unlike the garden SDK PR, this one isn't a temporary fix. We want to allowlist this content type for good (and maybe more content types in the future, like images).

I also pulled over the revised PR guidelines from the other repo while I was here.

## Testing

Tested with the code in https://github.com/Garden-AI/garden/pull/80. I applied the Terraform changes, ran the SDK code that adds the content-type header, and confirmed we get an unmangled model.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,4 +1,4 @@
-Fixes # (Some reference to an issue)
+Link to any related issues, companion PRs, or blockers. eg, `Fixes #37`
 
 ## Overview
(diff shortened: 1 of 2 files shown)


OwenPriceSkelly (approved):
love getting to review a terraform PR I understand. looks good to me!

PR #20, opened 2023-06-12
Add route for updating record on Datacite

isaac-darling:
Component of [#140](https://github.com/Garden-AI/garden/issues/140)

## Overview

Adds a route to the Garden backend for making PUT requests to Datacite.

## Testing

TBD.

### DIFF
--- a/app/doi.py
+++ b/app/doi.py
@@ -58,7 +58,7 @@ def call_datacite(event, _context, _kwargs):
         # DataCite successful response *would* have all our repo info,
         # so extract just the newly minted DOI for the response body
         if method == "POST":
-            return return_response.update(
+            return_response.update(
                 {
                     "body": json.dumps(
                         {
@@ -67,5 +67,4 @@ def call_datacite(event, _context, _kwargs):
                     )
                 }
             )
-        else:
-            return return_response
+        return return_response
--- a/app/doi.py
+++ b/app/doi.py
@@ -95,4 +95,4 @@ def update_metadata(event, _context, _kwargs):
     else:
         # DataCite successful response *would* have all our repo info,
         # so extract just the newly minted DOI for the response body
-        return {"statusCode": 201}
+        return {"statusCode": 200}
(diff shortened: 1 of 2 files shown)


app/doi.py
```diff
@@ -55,3 +55,44 @@ def mint_doi(event, _context, _kwargs):
...
+        # failed to set prefix due to malformed payload
+        return {"statusCode": 400, "body": str(e)}
+    except HTTPError as e:
+        # DataCite error responses seem safe to include outright
+        # propagate errors from requests.raise_for_status directly
+        return {"statusCode": 500, "body": str(e)}
+    else:
+        # DataCite successful response *would* have all our repo info,
+        # so extract just the newly minted DOI for the response body
+        return {"statusCode": 201}
```
WillEngler, line 98: It seems like this is the same as the existing function, except with post vs put and returning the resource in the body vs not. Can we DRY this?

WillEngler, line 98: Also, I don't think 201 Created is appropriate if the DOI already existed.

WillEngler (changes requested):

app/doi.py
isaac-darling: I asked Owen about how to approach this and he gave the opposite input, favoring readability.

isaac-darling: Fixed!

WillEngler: Haha, ok I disagree with Owen but if you have his blessing, 🤷

isaac-darling: Done!

app/doi.py
```diff
@@ -47,11 +57,15 @@ def mint_doi(event, _context, _kwargs):
     else:
         # DataCite successful response *would* have all our repo info,
         # so extract just the newly minted DOI for the response body
-        return {
-            "statusCode": 201,
-            "body": json.dumps(
+        if method == "POST":
+            return return_response.update(
```
OwenPriceSkelly, line 61: pitfall I personally stumble into constantly: python's dict.update is inplace, so as-is this would return None
```suggestion
            return_response.update(...)
        ...
        return return_response
```

OwenPriceSkelly (approved):
lgtm!

Issue #18, opened 2023-06-09
Add backend endpoint and buckets for presigned URL model upload


WillEngler:
Create a dev and prod S3 bucket with the right permissions attached to the app lambda. Create a new endpoint like POST /presigned-model-url in the app lambda. It should take a qualified model name like "willengler@uchicago.edu/example_model" and return a presigned URL to upload/download from S3. 

# Assumptions:
1.
2.

# Acceptance Criteria
Given I am a CLI user, when the CLI POSTs to {garden_url}/presigned-model-url with the fully qualified name of the model I want to upload/download, then I receive a presigned URL that I can use to upload my model with boto.

WillEngler:
Closed by #21

PR #21, opened 2023-06-13
Add new S3 buckets and presigned URL endpoint

WillEngler:
Implements #18

## Overview

Terraform changes:
- Adds dev and prod model storage buckets.
- Gives our app Lambda full read/write access to them.

API changes:
- Adds POST /presigned-url route that can generate presigned URL and field metadata for uploading/downloading zipped MLFlow MLModel directories.

I've already executed the terraform, so the new buckets exist.

## Discussion

For now I'm just using the dev bucket unconditionally. We can revisit selecting dev vs prod with the broader dev vs prod tickets on the backlog.

## Testing

I used the test harness to do basic testing of happy and error paths. I also plugged an upload URL generated from the harness into my CLI prototype to confirm that it correctly generated the upload URL.


app/presigned_urls.py
```diff
@@ -0,0 +1,89 @@
...
+    except Exception as e:
+        return {"statusCode": 500, "body": str(e)}
+
+    return {
+        "statusCode": 200,
+        "body": json.dumps(url_and_fields_payload),
+    }
+
+
+def is_probably_valid_object_path(object_path: str):
```
OwenPriceSkelly, line 70: Just to make sure that I'm on the same page, we don't really mind if something less-obviously malformed slips through because we think it's unlikely, or because it's fine to always just return 500 plus whatever exception message from the s3 client anyways?

OwenPriceSkelly (approved):
lgtm! also, shoutout to chatgpt for answering my terraform questions so you don't have to

app/presigned_urls.py
WillEngler: I'm of a mind right now where I'm letting our backend be "dumb pipes" and have the serious business logic live in the CLI. When we get to the "real" backend I want to invert that and have our serious validation on the backend. But yeah for now the reasoning is that I'm trusting the CLI to do its job so I think something malformed slipping through is unlikely. (And if it's super badly malformed then S3 would rebel)

PR #22, opened 2023-06-13
add route for publishing garden

### LINKED ISSUE #15: Globus Search Index Ingest backend route
The Garden client currently is writing directly to the Globus Search Index, which requires every Garden user to have write permissions.  

Task:  implement a route on the Garden backend that writes to the Garden search index.
It can be based on https://github.com/Garden-AI/globus-search-writer-lambda (which is actually in use for DLHub, but lives in the Garden-AI github.)  However, since routes in the Garden backend app are already authorized by Globus Auth, this backend route might be simpler, depending on how much policy we want to implement.

isaac-darling:
`Fixes #15`

## Overview

Add the ability to hit the backend to publish a garden.

## Discussion

Has some questions for reviewers left in the code comments.


WillEngler:
We had a Zoom sync on this. Some of the big things we discussed:
- Handling different error cases
- Making sure not to reference items in the auth lambda. (It won't be included in the app lambda's deployment bundle)
- Doing basic testing with `python local_test_harness.py`

app/lambda_function.py
```diff
@@ -1,25 +1,111 @@
...
+    }
+
+    publish_result = search_client.create_entry(GARDEN_INDEX_UUID, gmeta_ingest)
+
+    max_intervals = 30
+    task_result = search_client.get_task(publish_result["task_id"])
+    while task_result["state"] not in {"FAILED", "SUCCESS"}:
+        if not max_intervals:
+            return {
+                "statusCode": 408,  # there isn't a perfect analog here, so perhaps a simple 500 may even be better
```
WillEngler, line 84: imo this is fine and 500 would also be fine, so whatever you think

app/lambda_function.py
```diff
@@ -1,25 +1,111 @@
...
+                "body": json.dumps(
+                    {
+                        "message": f"Server timed out waiting for publish task to finish, you can manually check its progress with the task id: {task_result['task_id']}"
+                    }
+                ),
+            }
+        sleep(2)  # suitable wait time and timeout duration?
```
WillEngler, line 91: Picking these kinds of numbers is always kind of guesswork. 1 second interval would be ok too. If we had an exponential backoff generator lying around I'd use that :) 

For now I think this is ok and we can easily tweak if it's causing a problem.

app/lambda_function.py
```diff
@@ -1,25 +1,111 @@
...
-                'message': 'Hello there. You must be an authenticated Globus user.',
-            })
+                "message": "Hello there. You must be an authenticated Globus user.",
+            }
+        ),
     }
 
 
+@app.route("/garden-search-record", methods=["POST"])
+def publish(event, context, kwargs):
```
WillEngler, line 48: Can we pull the `publish` implementation and `get_secret` into a separate file? I'd like to keep this main file mostly just the routing setup.

WillEngler (approved):
I have one ask, but otherwise LGTM!

PR #23, opened 2023-06-14
add some documentation for how to create scopes, and all current Gard…

WillEngler:
For Max's work on running integration tests in GHA, we were running into problems using the ConfidentialAppAuthClient with the existing garden action_all scope.

```
confidential_client = globus_sdk.ConfidentialAppAuthClient(CLIENT_ID, CLIENT_SECRET)
authorizer = globus_sdk.ClientCredentialsAuthorizer(confidential_client, "https://auth.globus.org/scopes/0948a6b0-a622-4078-b0a4-bfd6d77d65cf/action_all")
```

led to 

```
ValueError: Attempting get new access token for client credentials authorizer didn't return exactly one token. Ensure scopes https://auth.globus.org/scopes/0948a6b0-a622-4078-b0a4-bfd6d77d65cf/action_all are for only one resource server.
```

I think this is because the garden_all scope takes a Globus Groups scope as a dependent scope. This was just copypasta from an action provider tutorial. We don't actually need that scope. So I made a new garden scope with no dependent scope and dashed off some quick notes so that we know where to find the docs on making scopes and what our scopes are.


WillEngler:
Going to cavalierly merge without review because this is just adding some useful documentation for the benefit of the dev team and not altering any code.

PR #25, opened 2023-06-15
fix requested scope in publish route

isaac-darling:


WillEngler (approved):
Thanks for the quick turnaround, LGTM

PR #30, opened 2023-06-16
align with the SDK and use doi subjects

isaac-darling:
## Overview

When the SDK switches to expecting DOI subjects, the publish route needs to accommodate. This is that fix.

## Testing

Full testing can be performed prior to merge after updates to the SDK.


WillEngler (approved):
LGTM

Issue #19, opened 2023-06-09
Clean up model artifact resources


WillEngler:
Once we have the S3 presigned URL approach working, clean up the artifact repository. Remove it from the terraform code and our actually-deployed AWS resources.

WillEngler:
Closed by #31

PR #31, opened 2023-06-30
Remove the MLFlow backend deployment

WillEngler:
Implements #19 

## Overview

Removes the backend resources for the MLFlow server.

## Discussion

In addition to these Terraform-managed resources, I removed the manually provisioned VPC that the ECS app was running within.

## Testing

I have executed the Terraform and done some smoke tests to confirm that backend operations are still working.


OwenPriceSkelly (approved):
we should keep a leaderboard for negative diffs. LGTM 🫡

PR #32, opened 2023-06-30
add unit tests and cleanup handling of secrets

isaac-darling:
## Overview

Adds unit tests where there were previously none and reconfigures the way the repo handles secret management.

## Discussion

Currently several placeholders, awaiting deployment on AWS to fill in hard-coded strings.

## Testing

Testing will be performed after the placeholders are removed.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,43 @@
+name: tests
+on:
+  push:
```
WillEngler, line 3: Does this mean that it will run on every push to any branch? If so, is it possible to get it to run only on PRs? https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request

app/requirements.txt
```diff
@@ -1,2 +1,5 @@
-requests==2.28.2
-urllib3<2
\ No newline at end of file
+setuptools>=65.5.1
+requests>=2.20.0
+urllib3<2
+boto3
+globus_sdk
```
WillEngler, line 5: Could you pin a version of globus-sdk? (We want to be broad in our requirements for the SDK but the backend can/should be as specific as possible)

Also, how was this working before we had globus_sdk as a requirement!?!

app/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    # Create a Secrets Manager client
+    session = boto3.session.Session()
+
+    client = session.client(service_name="secretsmanager", region_name=region_name)
+
+    get_secret_value_response = client.get_secret_value(SecretId=secret_name)
+    try:
+        return json.loads(get_secret_value_response["SecretString"])
+    except json.JSONDecodeError:
+        return get_secret_value_response["SecretString"]
```
WillEngler, line 19: Interesting, do we see both cases in the wild? Do different environments get different response formats? This is maybe worth a comment.

authorizer/lambda_function.py
```diff
@@ -17,7 +18,7 @@ def get_secret():
     get_secret_value_response = client.get_secret_value(
         SecretId=secret_name
     )
-    return eval(get_secret_value_response['SecretString'])
+    return json.loads(get_secret_value_response['SecretString'])
```
WillEngler, line 21: nice

app/test_app.py
```diff
@@ -0,0 +1,64 @@
...
+        json.dumps({"direction": "upload", "s3_path": "willengler@uchicago.edu/example-model/model.tar"}),
+        json.dumps({"direction": "upload", "s3_path": "willengler@uchicago.edu/example-model/model.zip"}),
+        json.dumps({"direction": "download", "s3_path": "willengler@uchicago.edu/example-model/model.zip"}),
+    ]
+    events = [{"path": "/presigned-url", "httpMethod": "POST", "body": payload} for payload in payloads]
+
+    assert (res:=lambda_handler(events[0], None))["statusCode"] == 400 and "direction" in json.loads(res["body"])["message"]
+    assert (res:=lambda_handler(events[1], None))["statusCode"] == 400 and "format" in json.loads(res["body"])["message"]
+    assert (res:=lambda_handler(events[2], None))["statusCode"] == 200 and json.loads(res["body"])  # make sure return body is valid json
+    assert (res:=lambda_handler(events[3], None))["statusCode"] == 200 and json.loads(res["body"])
```
WillEngler, line 53: Maybe my first time seeing the walrus operator in the wild.

app/test_app.py
```diff
@@ -0,0 +1,64 @@
+import json
+import pytest
+
+from lambda_function import lambda_handler
+from tiny_router import RouteNotFoundException
+
+
+def test_sanity() -> None:
```
WillEngler, line 8: hahahah

WillEngler (commented):
Good job on the test cases and fiddling with everything to get it working.

app/utils.py
isaac-darling: we do, I added a comment about it. namely, Eric's original secret has that format

app/requirements.txt
isaac-darling: there was nothing to "not work" before, since the tests were just added :)

.github/workflows/ci.yaml
isaac-darling: according to the docs, my change will have it run tests any time a PR is opened, reopened, or has its head updated.

WillEngler (approved):

PR #33, opened 2023-07-05
do not attempt json decode secrets every time

isaac-darling:
## Overview

Fixes unintended json string decode bug.


WillEngler (approved):
LGTM, thanks

PR #35, opened 2023-07-06
add support for batching

isaac-darling:
## Overview

Improves wait time on client side by handling several presigned_url tasks with a single http request.

## Testing

This will need to go live in order to fully test the corresponding [client PR](https://github.com/Garden-AI/garden/pull/198).

### DIFF
--- a/app/presigned_urls.py
+++ b/app/presigned_urls.py
@@ -63,7 +63,7 @@ def make_presigned_url(event, _context, _kwargs):
         except Exception as e:
             return {"statusCode": 500, "body": str(e)}
 
-        responses.append({"body": json.dumps(url_and_fields_payload)})
+        responses.append(url_and_fields_payload)
 
     # provides all of the responses when successful, otherwise only the first error encountered
     return {"statusCode": 200, "body": json.dumps({"responses": responses})}
--- a/app/presigned_urls.py
+++ b/app/presigned_urls.py
@@ -59,11 +59,13 @@ def make_presigned_url(event, _context, _kwargs):
 
         make_url = url_makers[direction]
         try:
+            response_payload = {"model_name": s3_path[:s3_path.find("/model.zip")]}
             url_and_fields_payload = make_url(s3_path, bucket_name)
+            response_payload.update(url_and_fields_payload)
         except Exception as e:
             return {"statusCode": 500, "body": str(e)}
 
-        responses.append(url_and_fields_payload)
+        responses.append(response_payload)
 
     # provides all of the responses when successful, otherwise only the first error encountered
     return {"statusCode": 200, "body": json.dumps({"responses": responses})}
--- a/app/presigned_urls.py
+++ b/app/presigned_urls.py
@@ -40,6 +40,7 @@ def make_presigned_url(event, _context, _kwargs):
     payload = json.loads(event["body"])
     batch = payload["batch"]
     direction = payload["direction"]
+    responses = []
 
     if direction not in url_makers:
         return {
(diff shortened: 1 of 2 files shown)


app/presigned_urls.py
```diff
@@ -38,33 +38,40 @@ def make_download_url(full_object_path: str, bucket: str) -> dict:
...
-            "body": json.dumps({"message": message}),
+    for i, s3_path in enumerate(batch):
+        if not is_probably_valid_object_path(s3_path):
+            message = "'s3_path' must be formatted like '<email address>/<model short name>/model.zip'. "
+            message += f"Got {s3_path}"
+            batch[i] = {
+                "statusCode": 400,
+                "body": json.dumps({"message": message}),
+            }
+            break
```
WillEngler, line 58: When we hit an error (here and on like 65) can we return early with the error instead of breaking?

app/test_app.py
```diff
@@ -40,15 +40,16 @@ def raise_for_status(self) -> None:
...
+        json.dumps({"direction": "upload", "batch": ["willengler@uchicago.edu/example-model/model.zip"]}),
+        json.dumps({"direction": "download", "batch": ["willengler@uchicago.edu/example-model/model.zip",
+                                                       "willengler@uchicago.edu/sample-model/model.zip"]}),
     ]
     events = [{"path": "/presigned-url", "httpMethod": "POST", "body": payload} for payload in payloads]
 
     assert (res:=lambda_handler(events[0], None))["statusCode"] == 400 and "direction" in json.loads(res["body"])["message"]
-    assert (res:=lambda_handler(events[1], None))["statusCode"] == 400 and "format" in json.loads(res["body"])["message"]
+    assert (res:=lambda_handler(events[1], None))["statusCode"] == 400 and "format" in json.loads(res["body"])["responses"][0]["body"]
     assert (res:=lambda_handler(events[2], None))["statusCode"] == 200 and json.loads(res["body"])  # make sure return body is valid json
```
WillEngler, line 53: Could you add a case where you verify the full shape of the return value?

app/presigned_urls.py
```diff
@@ -38,33 +38,40 @@ def make_download_url(full_object_path: str, bucket: str) -> dict:
...
-        message = "'s3_path' must be formatted like '<email address>/<model short name>/model.zip'. "
-        message += f"Got {s3_path}"
-        return {
-            "statusCode": 400,
-            "body": json.dumps({"message": message}),
+    for i, s3_path in enumerate(batch):
+        if not is_probably_valid_object_path(s3_path):
+            message = "'s3_path' must be formatted like '<email address>/<model short name>/model.zip'. "
+            message += f"Got {s3_path}"
+            batch[i] = {
```
WillEngler, line 54: We should avoid mutating/overwriting our input array. Like make a separate output array and append to it, and then we don't even have to explicitly keep track of indices

app/presigned_urls.py
```diff
@@ -38,33 +38,40 @@ def make_download_url(full_object_path: str, bucket: str) -> dict:
...
+
+        make_url = url_makers[direction]
+        try:
+            url_and_fields_payload = make_url(s3_path, bucket_name)
+        except Exception as e:
+            batch[i] = {"statusCode": 500, "body": str(e)}
+            break
+
+        batch[i] = {
+            "statusCode": 200,
```
WillEngler, line 68: I don't think we need to return a statusCode for each URL

WillEngler (changes requested):
Requesting some code quality and interface changes. Good progress though

WillEngler (approved):
Looking very clean, great work

PR #40, opened 2023-11-09
Remove presigned URL route

WillEngler:
Opportunistic cleanup. Forgot to remove this route the other week when we removed the SDK code that invokes it.

## Overview

Removes all code related to proxying MLFlow bundle uploads and downloads.

## Discussion

Thank you for your service, presigned URLs.

## Testing

Ran unit tests as a sanity check that I didn't break the app. This just surgically takes out the /presigned-url routes


OwenPriceSkelly (approved):
lgtm!

PR #44, opened 2023-11-27
Add new notebook buckets and route that writes to them

WillEngler:
First part of https://github.com/Garden-AI/garden/issues/345

## Overview

*Infra*
This PR removes the old cruft model buckets and adds new buckets to store .ipynb files.

*Backend app*
This PR also adds a `POST /notebook` route that lets the CLI/SDK upload notebook JSON and get back a public URL where people can GET the notebook JSON later.

## Discussion

*Infra*
I also parameterized the S3 module. There was some weird consequences to two separate terraform states "owning" the same resources.

## Testing

I ran `terraform apply` on both dev and prod. I tested the new route locally and confirmed that it sent me back a valid public URL. I also added a simple unit test for the new route.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,6 +30,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          pip install --upgrade pip
           python -m pip install -r ./app/requirements.txt && python -m pip install -r ./app/test-requirements.txt
 
       - name: Test with pytest
--- a/app/lambda_function.py
+++ b/app/lambda_function.py
@@ -2,6 +2,7 @@ import json
 
 from doi import call_datacite
 from search_record import publish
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
 
       - name: Install dependencies
         run: |
+          pip install --upgrade pip
```
WillEngler, line 33: This is because of a blocking library vuln check complaining about something that can happen when you pip install from a mercurial repo on old versions of pip 🙄

OwenPriceSkelly (approved):
for sale: `prod_models`, never used 

lgtm!

WillEngler:
> for sale: `prod_models`, never used
> 
> lgtm!

lmao

PR #45, opened 2023-11-28
Dump notebook JSON to string before uploading it to S3

WillEngler:
When testing locally I got confused about what the framework does and doesn't serialize/deserialize. That led to a Python dict getting passed to S3 upload instead of a JSON string. This PR fixes that.


WillEngler:
Nevermind, it's better to change what's going on from the client side

PR #55, opened 2024-02-29
Delete garden search record

WillEngler:
Part of #421

## Overview

This PR adds a DELETE variant to /garden-search-record. Specify the DOI of the garden you want removed, and it will do so.

## Discussion

There is deliberately still no access control. Holding off on that until the Real App.

## Testing

I did some manual testing in dev - farewell MaxTestGarden3. Also added a unit test for the new route.

### DIFF
--- a/app/lambda_function.py
+++ b/app/lambda_function.py
@@ -1,7 +1,7 @@
 import json
 
 from doi import call_datacite
-from search_record import publish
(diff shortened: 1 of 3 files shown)


OwenPriceSkelly (commented):
LGTM!

PR #69, opened 2024-04-02
Revamp/hello auth

### LINKED ISSUE #65: Implement new backend routes: Globus auth
globus auth for new app instead of auth lambda 

Acceptance criteria:
- parity with the current auth lambda -- the hello-world app only greets requests from users who are logged in to an active globus account

OwenPriceSkelly:
closes #65 

## Overview

This PR starts fleshing out the hello-world app by implementing auth. Most of the auth code was lovingly plagiarized from globus compute's fastapi app. 

Now, hitting the "/greet" route with will greet you personally if the bearer token indicates that you're authenticated with globus auth and have the Garden `action_all` scope, otherwise it will turn you away with an appropriate error. 

I also made some early organizational decisions about the directory structure of the app, which I cover in the readme. 

## Testing

Added some unit tests (and tweaked ci.yml to actually run them) and tested against the local dev server with and without a valid bearer token. 

In this way, I also learned that the sdk doesn't currently request the `action_all` scope for Garden (just the test scope). I have a mini PR over there that changes that but until then the token in your `tokens.json` won't be enough to earn you a bespoke greeting if you test it yourself.


garden-backend-service/README.md
```diff
@@ -0,0 +1,54 @@
...
+│   ├── auth              # modules that don't `import fastapi`
+│   │   ├── __init__.py
+│   │   ├── auth_state.py
+│   │   └── globus_auth.py
+│   ├── config.py         # settings/reads env vars
+│   └── main.py           # inits app/imports routers 
+└── tests       # you guessed it
+    ├── __init__.py
+    └── test_routes.py
+```
```
WillEngler, line 30: This layout seems great

garden-backend-service/README.md
```diff
@@ -0,0 +1,54 @@
...
+Running `./run-dev-server.sh` will spin up the app at http://localhost:5500 in a container almost exactly like the one used for deployment, with the following tweaks:
+
+- mounts ./src as shared volume (instead of COPY) so local changes propagate into the container
+- also mounts ./.env file so the app can read e.g. API_CLIENT_SECRET vars. The .env file does not exist in the (public) image we deploy from.
+- runs uvicorn with --reload so changes you make in ./src get picked up immediately
+
+In addition to curl, postman etc you can also visit http://localhost:5500/docs in a browser for interactive docs that let you exercise specific endpoints.
+
+Unit tests live in `./tests/` and are run by github actions on PRs, or can be run locally w/ `poetry run pytest`
+
```
WillEngler, line 51: This rocks, big improvement over the old devex

garden-backend-service/src/api/dependencies/auth.py
```diff
@@ -0,0 +1,34 @@
...
+from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
+
+from src.auth.auth_state import AuthenticationState
+
+
+logger = logging.getLogger("app")
+
+
+def _get_auth_token(
+    authorization: HTTPAuthorizationCredentials = Depends(HTTPBearer(auto_error=False)),
```
WillEngler, line 12: Learning about FastAPI dependency injection from this PR, seems very nice to work with

garden-backend-service/src/api/dependencies/auth.py
```diff
@@ -0,0 +1,34 @@
...
+
+def _get_auth_state(
+    token: str = Depends(_get_auth_token),
+):
+    """Get an AuthenticationState object from the token in the Authorization header."""
+    return AuthenticationState(token)
+
+
+def authenticated(
+    auth_state: AuthenticationState = Depends(_get_auth_state),
```
WillEngler, line 28: It seems like there's a chain of dependencies here. `authenticated` needs `_get_auth_state`, and `_get_auth_state` needs `_get_auth_token`. But it doesn't seem like the intermediate private ones (`_get_auth_state` and `_get_auth_token`) are used independently anywhere. Is this something framework-y that you need to do or have you just chosen to break it up this way?

garden-backend-service/src/auth/auth_state.py
```diff
@@ -0,0 +1,71 @@
...
+    user. It is the "auth_state" object for the application within the context of a
+    request, showing "who" is calling the application (e.g. identity_id) and some
+    information about "how" the call is being made (e.g. scopes).
+
+    For the most part, this should not handle authorization checks, to maintain
+    separation of concerns.
+
+    NB: This class was lifted almost verbatim from funcx web service's auth
+    code. It seemed sane enough, plus doing so will hopefully streamline
+    the Developer Experience of plagiarizing other pieces of their code.
```
WillEngler, line 24: Love it

garden-backend-service/src/auth/auth_state.py
```diff
@@ -0,0 +1,71 @@
...
+        self.garden_default_scope: str = settings.GARDEN_DEFAULT_SCOPE
+        self.token = token
+
+        self.introspect_data: t.Optional[globus_sdk.GlobusHTTPResponse] = None
+        self.identity_id: t.Optional[uuid.UUID] = None
+        self.username: t.Optional[str] = None
+        self.scopes: t.Set[str] = set()
+
+        if token:
+            self._handle_token()
```
WillEngler, line 39: Are there cases in our app where this is constructed without a token? Just on the lookout for any potential states we could cut out. (Then we wouldn't need `.is_authenticated` for example.)

garden-backend-service/src/auth/globus_auth.py
```diff
@@ -0,0 +1,41 @@
...
+    auth_data = client.oauth2_token_introspect(
+        token, include="identity_set,identity_set_detail"
+    )
+
+    if log:
+        introspect_detail = getattr(auth_data, "data", auth_data)
+        logger.debug(
+            "auth_detail",
+            extra={"log_type": "auth_detail", "auth_detail": introspect_detail},
+        )
```
WillEngler, line 25: Nice, glad you're including logs around this from the jump.

garden-backend-service/src/config.py
```diff
@@ -0,0 +1,27 @@
...
+    )
+
+    class Config:
+        case_sensitive = True
+        env_file = find_dotenv()
+
+
+@lru_cache()
+def get_settings() -> Settings:
+    return Settings()
```
WillEngler, line 24: Understanding check: the `find_dotenv` call is the thing worth caching here?

garden-backend-service/tests/test_routes.py
```diff
@@ -1,14 +1,43 @@
...
 
 client = TestClient(app)
 
 
-def test_greeting():
+@pytest.fixture
+def mock_auth_state():
+    # Mock auth state for authenticated user
+    mock_auth = MagicMock(spec=AuthenticationState)
+    mock_auth.username = "Monsieur Sartre"
```
WillEngler, line 15: I am not merely playing the role of software engineer. I'm exercising my radical freedom by reviewing this PR on a sunny Tuesday morning. I will it so.

WillEngler (approved):
LGTM! There are some things I'm curious about but nothing blocking

garden-backend-service/tests/test_routes.py
OwenPriceSkelly: 🥖🍷

garden-backend-service/src/config.py
OwenPriceSkelly: I was maybe a little cargo-culty in plagiarizing this from globus compute because its purpose wasn't 100% obvious to me either, but my take on it is that it's probably a good idea for everything to share the same `Settings()` instance instead of instantiating their own. the full-stack-fastapi-template repo I was also looking at [does something similar](https://github.com/tiangolo/full-stack-fastapi-template/blob/c947cb09e1d03084c5037e2ef69d58c1c0397960/backend/app/core/config.py#L121) with a top level variable it reuses everywhere instead of instantiating `Settings()` more than once. 

the pydantic base class we use here automatically reads environment variables (set by dotenv or otherwise), and my suspicion is that they got burned by some subtle interaction between the state of os.environ and the pydantic class here, but I didn't dig too deeply

garden-backend-service/src/api/dependencies/auth.py
OwenPriceSkelly: yeah I actually wanted to get your input on this, we should sync up and take a look at the globus compute code I lifted it from. They seemed to like having the more granular intermediate dependencies (at least for auth stuff) but I wasn't sure if we thought we'd need that granularity

OwenPriceSkelly: they do use all of the intermediate dependencies at least once somewhere iirc

PR #70, opened 2024-04-15
Revamp/deployment aws creds

OwenPriceSkelly:
the MVP (Minimum Viable PR) for manually testing out an approach for injecting aws credentials at deployment time, since lightsail won't be able to assume IAM roles


WillEngler (approved):
LGTM

PR #71, opened 2024-04-17
Revamp/doi routes

### LINKED ISSUE #59: Implement backend route for DOI minting/registration

OwenPriceSkelly:
closes #59 

## Overview

This PR translates the datacite code from the current app lambda to the new fastapi project for the `/doi` resources. 

Infra-wise, this also includes the terraform to define an IAM user for the respective dev/prod lightsail deployments and attach a policy for reading config variables from secretsmanager. I've set the secret values for both manually. 

The deployment script injects the appropriate user credentials (stored as GH secrets) as env vars so the `config.Settings` object can populate itself with the config it reads from the aws secret on startup. 

I also added some new submodule structure for pydantic models we'll use for request/response bodies in `src/api/schemas`. Because the datacite rest api had an openapi.yaml I could use to autogenerate pydantic models, I went ahead and did so -- see the `src/api/schemas/datacite/__init__.py` module for the ones we actually use and `datacite/_full_schema.py` for the auto-generated stuff. 

## Discussion
The config variables workaround felt a little odd, but we can't easily set persistent env vars across deployments through lightsail nor can lightsail assume IAM roles. Even though (woke) chatgpt doesn't consider it "best practices", best practices aren't a practical option right now. Maybe worth revisiting in the future but this is close enough that I'm not losing any sleep over it.

I touched on this in the readme, but for "deploying" the local dev container, you now would at least set the AWS environment variables in your `.env` for the `garden_lightsail_user_dev` IAM user. This means that the local instance would have the same config values as the live dev deployment at baseline, but if you want to override any config variables just put them in the `.env` too (values read from `.env` get priority over values read from secretsmanager).

## Testing

Primarily tested with a local dev server. Initially just by hand-crafting requests through the interactive docs, but I also sanity checked by pointing my garden CLI at localhost and confirmed that draft dois were minted with `garden create` and that dois were registered properly (on the datacite test api) with `garden register-doi`.


infra/modules/lightsail/main.tf
```diff
@@ -14,7 +14,27 @@ resource "aws_lightsail_container_service" "garden_service" {
 
 # IAM stuff
 # note: lightsail can't assume iam roles, so we define an iam user here to attach policies to
-# but inject the actual credentials for the app at deployment time with the gh action
+# but inject the actual user credentials for the app at deployment time with the gh action
 resource "aws_iam_user" "lightsail_user" {
   name = "garden_lightsail_user_${var.env}"
 }
+
+data "aws_iam_policy_document" "secrets_policy" {
```
OwenPriceSkelly, line 22: I had half a mind to put this in the secretsmanager module, but since it's so tightly coupled to the IAM user dedicated to the lightsail deployments I figured it made just as much sense here. Could be convinced either way

WillEngler: I like the conceptual grouping

WillEngler (approved):
Things are moving along! Great stuff

PR #72, opened 2024-04-18
Revamp/docker push token

### LINKED ISSUE #62: Implement backend route for making ECR token (for user docker images)

OwenPriceSkelly:
closes #62 
## Overview

This implements the /docker-push-token endpoint for the revamp. the new terraform is almost identical to the analogous terraform in the lambda module. 

I also added a few entries to the `garden-backend-env-vars/dev` and `/prod` secrets for the corresponding ECR repos and IAM roles to assume 

## Testing
Tested against a dev container locally but with the same credentials as the live dev instance to test that it was reading from secretsmanager and assuming the role correctly. I didn't actually try pushing a container or anything, just made sure that the request body coming back looked right with ECR session tokens where the sdk would expect them.

I even tossed in a unit test

### DIFF
--- a/garden-backend-service/src/api/schemas/docker.py
+++ b/garden-backend-service/src/api/schemas/docker.py
@@ -1,12 +1,9 @@
 from pydantic import BaseModel
-from src.config import get_settings
-
-settings = get_settings()
 
 
 class ECRPushCredentials(BaseModel):
     AccessKeyId: str
     SecretAccessKey: str
     SessionToken: str
-    ECRRepo: str = settings.ECR_REPO_ARN
+    ECRRepo: str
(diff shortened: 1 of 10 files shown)


garden-backend-service/src/api/schemas/docker.py
OwenPriceSkelly, line 1: If you're wondering: "how did you feel about hiding such a small class all the way out in its own module?" it did feel a little silly. but not enough to shake my commitment to the Project Structure

WillEngler: I think it's great! We could use more project structure fanaticism in the CLI repo.

WillEngler (approved):
LGTM!

PR #74, opened 2024-04-22
Revamp/garden search record

### LINKED ISSUE #60: Implement backend route for publish/delete search record

OwenPriceSkelly:
closes #60 

## Overview
This PR implements the `/garden-search-record/` routes to publish garden metadata to (and delete from) the appropriate globus search index. This means the revamped backend can officially do everything the old one can (🎉🥳🍾)

## Discussion
The only change in behavior vs. the old backend is that we now do the `is_doi_registered` check before actually deleting. 

This PR is the first place I've introduced the actual concept of a `Garden` or `Entrypoint` to a schema - specifically, I copied over any pydantic models we might need to validate a `PublishedGarden.json()` request body. 

I made an organizational judgement call to keep those buried as part of a `schemas.search` submodule (rather than enshrining them in something like `schemas.garden_ai`) since this is just the schema that the search index is going to use as long as it stays around. I.e. these are the schemas specifically for a "garden search record", not any garden in general. 

## Testing

manually with the CLI and local test instance of the app. skipped the unit test for this one since I don't think the `/garden-search-record/` routes will have a long enough shelf life to really justify it

### DIFF
--- a/garden-backend-service/src/api/dependencies/auth.py
+++ b/garden-backend-service/src/api/dependencies/auth.py
@@ -5,7 +5,7 @@ from fastapi.security import HTTPAuthorizationCredentials, HTTPBearer
 from src.auth.auth_state import AuthenticationState
 
 
-logger = logging.getLogger("app")
+logger = logging.getLogger()
 
 
 def _get_auth_token(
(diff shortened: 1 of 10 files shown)


WillEngler (approved):
LGTM!

PR #76, opened 2024-04-23
Revamp/bump pydantic

### LINKED ISSUE #75: revamp: use pydantic 2.x
I initially kept the same pydantic 1.x constraint for `garden-backend-service` that we have for the garden sdk, but now that I think about it a little harder we should bump the backend to 2.x while it's still easy to do so. They might eventually share a `garden-common` package or something that would need the versions to match up, but in the meantime there's no reason to constrain the backend unnecessarily

OwenPriceSkelly:
closes #75 

## Overview
Bumps pydantic to 2.x and fixes a few things that broke when I did so. 

## Discussion
I was able to introduce a surprising number of incompatibilities in just a few short weeks, so I'm glad I thought to do this early. Some gotchas we'll want to keep in mind when we do this for the sdk: 
  - their config mechanism has changed (to a special `model_config` attribute instead of a nested `class Config`) so fields that start with `model_*` will cause pydantic to complain that it's a protected namespace (this was easy to kludge for ModelMetadata) 
  - `unique_items` isn't supported for list fields anymore; this was more involved but I found a very clean workaround from a github issue that was easy to plagiarize. 

## Testing

Ran the unit tests and manually hit every route again from the CLI and/or interactive docs


garden-backend-service/src/api/schemas/search/_garden_sdk_schema.py
```diff
@@ -1,8 +1,29 @@
...
+    if len(v) != len(set(v)):
+        raise PydanticCustomError("unique_list", "List must be unique")
+    return v
+
+
+UniqueList = Annotated[
+    List[T],
+    AfterValidator(_validate_unique_list),
+    Field(json_schema_extra={"uniqueItems": True}, default_factory=list),
+]
```
WillEngler, line 26: Ooh fun

WillEngler (approved):
At long last, 2.x!

PR #86, opened 2024-05-02
Fix/remove trailing slashes

OwenPriceSkelly:
cowboy PR 🤠 to remove the (apparently unnecessary) trailing slashes from every route. 

I preferred this over just adding trailing slashes everywhere in the SDK because a) it seems like a cleaner convention and b) it was causing a weird mismatch in behavior between a localhost deployment and the actual dev deployment.

if you sent a request with a missing slash to localhost, it was happy to redirect:
> INFO:     192.168.65.1:51974 - "POST /doi HTTP/1.1" 307 Temporary Redirect
> INFO:     192.168.65.1:51974 - "POST /doi/ HTTP/1.1" 201 Created

but not lightsail: 
> INFO: 172.26.29.83:19894 - "POST /doi HTTP/1.1" 307 Temporary Redirect
> INFO: 172.26.29.83:19894 - "GET /doi/ HTTP/1.1" 405 Method Not Allowed

no idea why it changed from POST to GET, but seems like if we omit the trailing slashes everyone's happy.


garden-backend-service/src/api/routes/doi.py
```diff
@@ -32,9 +38,44 @@ async def mint_draft_doi(
     return {"doi": result.data.attributes.doi}
 
 
-@router.put("/", status_code=status.HTTP_200_OK)
+@router.put("", status_code=status.HTTP_200_OK)
 async def update_datacite(
-    body: datacite.Doi,
+    body: datacite.Doi = Body(
+        examples=[
```
OwenPriceSkelly, line 44: these examples should show up in the interactive docs for testing w/ more realistic payloads

garden-backend-service/src/api/routes/doi.py
```diff
@@ -32,9 +38,44 @@ async def mint_draft_doi(
...
+                            }
+                        ],
+                        "creators": [
+                            {
+                                "name": "Mendel, Gregor",
+                            }
+                        ],
+                        "titles": [
+                            {
+                                "title": "GPT Garden (Generative Pea-trained Transformers)",
```
OwenPriceSkelly, line 66: usually hate to revive a dead joke but I really needed to get this out of my system

garden-backend-service/src/api/routes/doi.py
OwenPriceSkelly: as opposed to the automatically generated datacite examples, which were overkill

garden-backend-service/src/api/routes/doi.py
WillEngler: 🌱

WillEngler (approved):
LGTM

PR #88, opened 2024-05-15
hello-postgres

### LINKED ISSUE #78: revamp: hello postgres
acceptance criteria: terraform infra/ dedicated module for postgres; local backend instance can integrate with a local postgres install; `/hello-database` route to smoke test db access

OwenPriceSkelly:
closes #78 
## Overview
This implements a new terraform module at `infra/modules/rds` to provision postgres instances for dev and prod. It provisions the database with (hopefully) sane defaults for everything it can, but I'd appreciate a non-chatgpt vote of confidence on the actual values I landed on for the deployment type, amount of memory etc. 

This also implements a minimal `/hello-database` route which just asks postgres for the current time in order to sanity check that the app can actually connect to the database.

The dev infra has already been `terraform apply`-ed for the sake of testing the deployment (I've also manually deployed to the dev container service from this branch), but prod has not. 

## Discussion

I also tweaked the secretsmanager module to work around a chicken/egg-like problem with the DB credentials -- I wanted the source of truth for the credentials to be the values in the `garden-backend-env-vars/{dev,prod}` secret, but terraform needed to know the same username and password to provision the db itself. So I just had the secretsmanager module provide the credentials as outputs that dev/main.tf and prod/main.tf can then feed into the new rds module.  

This worked fine for the credentials, but the stickier chicken/egg problem is that we don't know the db endpoint until it's already provisioned. I don't think that will change unless we replace the entire db, but it's worth noting that the `DB_ENDPOINT` env var in `garden-backend-env-vars/{dev,prod}` needs to be set manually. 

## Testing

browse on over to https://api-dev.thegardens.ai/hello-database for a _timely_ greeting


garden-backend-service/pyproject.toml
```diff
@@ -17,6 +18,9 @@ cachetools = "^5.3.3"
 boto3 = "^1.34.84"
 email-validator = "^2.1.1"
 pydantic-settings = "^2.2.1"
+sqlalchemy = "^2.0.30"
+asyncpg = "^0.29.0"
+psycopg2 = "^2.9.9"
```
WillEngler, line 23: At long last

garden-backend-service/src/api/dependencies/database.py
```diff
@@ -0,0 +1,20 @@
...
+from src.config import get_settings
+
+postgres_url = get_settings().SQLALCHEMY_DATABASE_URL
+engine = create_engine(postgres_url, echo=True)
+SessionLocal = sessionmaker(bind=engine)
+
+# lifted from funcx web service in case we decide we need it
+_async_url = postgres_url.replace("://", "+asyncpg://")
+_async_engine = create_async_engine(_async_url, echo=True)
+_SessionLocalAsync = sessionmaker(bind=_async_engine, class_=AsyncSession)
```
WillEngler: Have you played with this? Do we want to be using it by default or are there some cases where we want the sync session over the sync session?

infra/dev/main.tf
```diff
@@ -51,6 +51,14 @@ module "lightsail" {
...
 }
 
+module "rds" {
+  source = "../modules/rds"
+
+  env         = var.env
+  db_username = module.secrets_manager.db_username
+  db_password = module.secrets_manager.db_password
+}
+
```
WillEngler, line 61: What was your thinking on using RDS over Lightsail's DB option?

WillEngler (approved):
Looks solid. There are a few things that are a little surprising that I'd like to talk over

PR #89, opened 2024-05-17
Create compose.yaml for easy local dev setup

hholb:
## Overview

I created a docker compose file that encapsulates the database setup for local development. I think this will speed up the process of setting up a local environment quite a bit and avoid having to manually setup and configure postgres.

It spins up 2 containers, one running the API  built from the `Dockerfile`, and one running postgres database. The database container only listens on the docker network setup by docker compose so it is not accessible outside the local setup.

The database container needs two environment variables, POSTGRES_DB and POSTGRES_PASSWORD, which define the default database name, and the password of `postgres` user. Define these in `.env.postgres`.

I updated the README with the docker compose instructions and removed the manual postgres setup instructions.

## Testing

Manually with docker.


garden-backend-service/README.md
```diff
@@ -35,64 +35,59 @@ garden-backend-service/
...
-host    all             all             172.17.0.0/16           md5
+
+Visit http://localhost:5500/docs and behold!
+
+Docker compose maps ./src, ./tests, and ./.env into the container so you can edit files locally
+and see the changes immediately reflected in the running container.
+
+Tear down and cleanup the containers:
+``` sh
+docker compose down
```
OwenPriceSkelly, line 80: is there an easy way to have docker compose automatically clean up the containers (like regular `docker run --rm`) when the app container exits / on a Ctrl-C? if it's not easy I wouldn't worry about it

garden-backend-service/README.md
```diff
@@ -35,64 +35,59 @@ garden-backend-service/
...
     AWS_SECRET_NAME=garden-backend-env-vars/dev
     GARDEN_ENV=dev
     # see below
     DB_USERNAME="garden_dev"
     DB_PASSWORD="your_password"
-    DB_ENDPOINT="host.docker.internal"
-    
-where the AWS access key variables correspond to the `garden_lightsail_user_dev` IAM user (which has permission to read the AWS secret). If you provide any additional variables which are also present in the `garden-backend-env-vars/dev` secret, the one you set in the .env file will take priority. 
-
-### Local PostgreSQL setup
```
OwenPriceSkelly, line 57: love this diff 🟥 🟥 🟥

OwenPriceSkelly: definitely simpler than the postgres.app approach

garden-backend-service/README.md
OwenPriceSkelly: also feel free to edit/delete the `run-dev-server.sh` script now that we're replacing it with docker compose

garden-backend-service/src/config.py
```diff
@@ -58,7 +58,7 @@ class Settings(BaseSettings):
     def SQLALCHEMY_DATABASE_URL(self) -> str:
         return f"postgresql+asyncpg://{self.DB_USERNAME}:{self.DB_PASSWORD}@{self.DB_ENDPOINT}/garden_db_{self.GARDEN_ENV}"
 
-    model_config = SettingsConfigDict(env_file=_dotenv_path, case_sensitive=True)
+    model_config = SettingsConfigDict(env_file=_dotenv_path, case_sensitive=True, extra='ignore')
```
OwenPriceSkelly: I think this is sane either way but I'd just as soon keep the default here and use a different env file (`.postgres.env`, maybe?) for the postgres variables.

Thinking along the lines of keeping the local "deployment" as close to the real deployment as possible, we should reserve the existing `.env` just for providing/overriding environment variables the app actually knows about

garden-backend-service/compose.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - ./tests:/app/tests
+      - ./.env:/app/.env
+    entrypoint: bash -c "uvicorn src.main:app --host 0.0.0.0 --port 80 --reload"
+    depends_on:
+      dev-db:
+        condition: service_healthy
+
+  dev-db:
+    hostname: dev-db
+    image: postgres:latest
```
OwenPriceSkelly:
```suggestion
    image: postgres:16
```
afaik this is currently the same as `:latest` but the actual dbs are using postgres 16 so might as well pin it here

OwenPriceSkelly (approved):
This looks great! I'll admit to sleeping on `docker compose` before now but this is definitely easier to get up and running. 

I do think we should separate the postgres env vars into their own file but otherwise: LGTM!

garden-backend-service/README.md
hholb: I don't think there is a way to automatically remove the containers, but you can run `docker compose up --build` to force a rebuild of the images.

garden-backend-service/src/config.py
hholb: That makes sense, a separate .env for postgres is no problem.

PR #90, opened 2024-05-22
hello-alembic

### LINKED ISSUE #79: revamp: hello sqlalchemy
acceptance criteria: baseline sqlalchemy/alembic configuration with scaffolded repository structure for the backend app

OwenPriceSkelly:
closes #79 

## Overview

This gets us up and running with alembic and the minimum sqlalchemy model(s) for a `users` table, which the `/greet` endpoint now interacts with. 

## Discussion

This got a little bigger than expected, but is effectively a hello-world PR for a few related things:
  - alembic migrations locally and for future deployments
  - sqlalchemy 2.x-style orm models 
  - interacting with the db through fastapi dependencies 
  - unit tests/fixtures that mock an async db connection 

Not planning on touching the users table in the near future, it was just the simplest table to hit with an existing endpoint / simple fastapi `authed_user` dependency. first real "load-bearing" table is going to be entrypoints.

Where possible I continued to proudly plagiarize the funcx-web-service repo - the helper methods on the `Base` sqlalchemy model were particularly inspiring - but a number of superficial differences due to e.g. them being on sqlalchemy 1.x and mid-migration from sync to async meant slightly less of a free lunch for me this time. 

## Testing
Had to make some tweaks to @hholb's very handy docker compose setup so that I could test the alembic migration stuff against the postgres container. specifically worth mentioning that I changed the postgres config to expect a `POSTGRES_USER` in `.postgres.env` that matches the `DB_USER` in the app's `.env`, which is a little redundant but mirrors the real infra better.


garden-backend-service/migrations/env.py
```diff
@@ -0,0 +1,98 @@
...
+def run_migrations_online() -> None:
+    """Run migrations in 'online' mode."""
+
+    asyncio.run(run_async_migrations())
+
+
+if context.is_offline_mode():
+    run_migrations_offline()
+else:
+    run_migrations_online()
```
WillEngler, line 98: Is the migration ini script, this file, and other alembic scaffolding largely pulled from alembic docs? (If it's standard boilerplate I won't scrutinize this as much)

garden-backend-service/src/models/user.py
```diff
@@ -0,0 +1,12 @@
...
+from sqlalchemy.orm import Mapped, mapped_column
+
+from .base import Base
+
+
+class User(Base):
+    __tablename__ = "users"
+    id: Mapped[int] = mapped_column(primary_key=True)
+    username: Mapped[str]
+    identity_id: Mapped[UUID] = mapped_column(unique=True)
```
WillEngler, line 12: Is this all data that we derive from the Globus Auth token? If so, what do we gain from persisting this?

garden-backend-service/migrations/env.py
OwenPriceSkelly: yeah it's entirely boilerplate -- except for reading the sqlalchemy url from settings and importing the models it was totally generated by alembic

garden-backend-service/src/models/user.py
OwenPriceSkelly: Exactly yeah this is just what we get at minimum from the globus auth token, and my sense is that it's likely to be handy if/when we need to integrate with other globus stuff. Didn't read too deeply into it past the first sentence here  https://docs.globus.org/api/auth/specification/#using_identities

OwenPriceSkelly: no immediate plans to do anything but store it Just In Case

WillEngler (approved):

PR #93, opened 2024-05-29
revamp: tables for entrypoints and related metadata

### LINKED ISSUE #80: revamp: entrypoints table
acceptance criteria: sqlalchemy class/postgres table for entrypoints. should include integration tests to run against local/dev/prod instances of the DB as part of the deploy script

### LINKED ISSUE #82: revamp: tables for "related work"  metadata
following #80 and #81, we'll need respective sqlalchemy models/postgres tables for "related work" metadata like datasets or model metadata. 

acceptance criteria: 
  - we have whatever tables ready that we'll need to migrate the contents of the search index without losing any information

### LINKED ISSUE #92: complete "authors" table / relationships
Paring down scope creep from #80. An authors field consisting of a list of strings are relevant for entrypoints as well as metadata of related objects such as papers/datasets. 

rather than puzzling out the specifics of the many-many relationships between authors, entrypoints, and papers etc upfront (especially since we don't actually have any paper metadata in the system yet), I'd like to leave it as just a one-many relationship between entrypoints and authors for the short term, and flesh it out as needed later.

OwenPriceSkelly:
closes #80; closes #82 (also closes #92, now obsolete)

## Overview

This PR defines the models for the first "load bearing" tables in the database, starting with entrypoints.

## Discussion

To prevent the scope of this PR creeping much further than it already had, I set up naive one-many relationships for all of the list attributes that entrypoints currently have, namely:
  - datasets
  - model metadata
  - papers 
  - repositories
  - ~tags~
  - ~test functions~
  - ~authors~ (these are now just `ARRAY(String)`s) 

~With the exception of test_functions, any of these should probably eventually be modeled with many-many relationships instead of the naive one-many relationships I need to reach "local_data parity". I could be persuaded to just bite the bullet here and set up the association tables for all of these from the beginning, but I don't think it's any more work to kick that down the road a little. Especially for e.g. author names (see #92), which we might want to start parsing as something more than just a plain string before relating the authors table to everything with an authors field. **Any input on my approach here would be very welcome 🙏**~
^Agreed with Will's comment, I was overthinking it and one-to-many thankfully does make the most sense for these.

**This PR also omits steps as a concept**. Current plans are to find time to meet with Logan and possibly some of our MIT collaborators to pick their brains about how Garden can help them enable code reuse in a way that gels better with our execution model. If the people want steps, we'll implement them as their own thing rather than tying them to entrypoints.

## Testing

Good news/bad news. Good is that the simple POST, GET and DELETE /entrypoint routes I defined work beautifully against the local postgres instance spun up with docker compose, manually tested with a json dump of all of Ryan Jacob's recent entrypoints (plus some dummy dataset/paper metadata on a few of them). 

Bad news is that implementing usable unit tests got so unwieldy that I'm just skipping them for now -- see #94


WillEngler (approved):
Looks like you're doing all the right things to me!

I also think one-to-many might just be correct here. I think I see what you're thinking, in that multiple entrypoints might reference the same paper, say. But I don't think we really want to do a thing where we link the same Paper entity to different users' entrypoints. Then what if one user changes the citation in a way the other user doesn't like? For that reason I think it's appropriate for these to be duplicated. Some far off future enhancement could make a "canonical" representation of a paper or dataset from a 3rd party information source, and then many to many would make more sense.

WillEngler (approved):
LGTM!

Issue #94, opened 2024-06-07
enhancement: async postgres(-like) DB fixtures for unit tests  


OwenPriceSkelly:
Working on #93, I've spent enough time banging my head into pytest walls that I'm pulling it out into its own issue. 

# Goals: 
  1. the DB fixture should be lightweight / not impose a significant performance penalty. 
    - Less concerned with local development than I am with slowing down the CI pipeline. We've got a pretty ergonomic setup for local testing against a proper postgres instance already with docker compose etc, I just want unit tests for coverage's sake.  
  2. The fixture should "just work" with our async code / tests.
  3. The fixture should "just work" with the postgres SQLAlchemy dialect  
  4. The fixture should target the fastAPI dependency injection system -- where possible I think we want to avoid mocking things at the level of individual routes. 

The current approach uses an in-memory SQLite DB by setting `mock_settings.SQLALCHEMY_DATABASE_URL = "sqlite+aiosqlite:///:memory:"`. 
### the Good about the current approach:
  - sqlite+aiosqlite is lightweight and async
  - The fixtures I've hacked together so far work by overriding the `get_settings` dependency and the `get_db_session` dependency, which helps keep the tests exercising the routes pretty simple. 
  
tl;dr we've currently got 1, ~2, and 4 from above 
### the Bad: 
  - doesn't play nicely when we use postgres-specific features (e.g. the `ARRAY(String)` type we're using for tags) 
  - A reasonable attempt to hack the postgres-specific array type ([gist](https://gist.github.com/OwenPriceSkelly/a7c30cb9d1cfded6ae9686932e03e3f5)) causes weird async/io errors. I ran into similar errors in the actual implementation, but those errors were fixed by setting the cascade settings correctly for entrypoint relationships. 
    - I'm not sure if these new errors means there's some kind of problem with the fake array type directly or if there's some inconsistency with the async tests / fixture database (of course also possible that I didn't *really* fix the real code, but it works fine with real data against a local postgres instance so I don't think that's the case).

OwenPriceSkelly:
I think it may be worth looking into pytest-postgresql as a possibility here if it doesn't slow down our tests too much. It seems like it relies on postgres being installed wherever the tests are run, which might be a dealbreaker depending on how that works with our GH actions

Issue #81, opened 2024-04-30
revamp: gardens table 


OwenPriceSkelly:
following #80, we'll also need a `gardens` sqlalchemy class/ postgres table

PR #96, opened 2024-06-12
revamp: gardens table

OwenPriceSkelly:
completes #81 (!)
## Overview

This PR defines the `"gardens"` table, which is the last table we need to persist everything currently on the search index 🎉 

It also implements basic `{POST,GET,DELETE} /garden` api routes.

## Discussion

This also configures the many-many relationship between gardens and entrypoints. The associative table `"gardens_entrypoints"` is keyed on respective DOIs instead of just their `id` fields. I opted for this because DOIs are a) unique and b) have meaning in the outside world, and I liked the idea of "baking in" that meaning to the system. Referential integrity felt like a nice way to guarantee we're not committing DataCite Sins 😇 

The many-many relationship isn't bidirectional, however -- I thought that giving entrypoints direct references to the gardens they appear in was a more complex data model and not something we *need* (yet?). 

I also wanted it to throw a 404 when trying to create a garden that references a non-existent entrypoint, which makes more sense when only gardens refer to entrypoints but not the other way around. 

The difference in behavior here is that deleting a Garden also deletes any corresponding rows from the `"gardens_entrypoints"` table, but deleting an Entrypoint will fail if it is still referenced by any Gardens. Only "orphan entrypoints" can be deleted. 

If we made the relationship bi-directional, the default would be that either could be deleted and references would update automatically. 

## Testing

I hacked together a little script ([gist](https://gist.github.com/OwenPriceSkelly/802cd4491443bfa1e2d72480afa1bafd)) to scrape the search index and test that POST/DELETE for gardens/entrypoints work as expected. 

I also included some test fixtures (specifically, some gardens that share a reference to the same entrypoint) and "stub" gpt-generated unit tests, but I didn't spend any time trying to make them work until we get around to #94.


garden-backend-service/src/models/garden.py
```diff
@@ -0,0 +1,37 @@
...
+    doi_is_draft: Mapped[bool] = mapped_column(default=True)
+    authors: Mapped[list[str]] = mapped_column(postgresql.ARRAY(String))
+    contributors: Mapped[list[str]] = mapped_column(postgresql.ARRAY(String))
+    tags: Mapped[list[str]] = mapped_column(postgresql.ARRAY(String))
+    description: Mapped[str | None]
+    publisher: Mapped[str]
+    year: Mapped[str]
+    language: Mapped[str]
+    version: Mapped[str]
+    entrypoint_aliases: Mapped[dict[str, str]] = mapped_column(postgresql.JSON)
```
WillEngler, line 30: Just noting, something that I don't see is an "owner" field or similar that we'd use in the routes to make sure that the creator of the garden is the only one who can edit/delete it. (I think it makes sense to take this in stages, so nothing wrong with getting in all of this functionality first. Just making sure this is on your radar.)

garden-backend-service/src/models/_associations.py
```diff
@@ -0,0 +1,10 @@
+from sqlalchemy import Table, Column, ForeignKey
+from src.models.base import Base
+
+
+gardens_entrypoints = Table(
+    "gardens_entrypoints",
+    Base.metadata,
+    Column("garden_id", ForeignKey("gardens.id"), primary_key=True),
+    Column("entrypoint_id", ForeignKey("entrypoints.id"), primary_key=True),
```
WillEngler, line 9: yayy

WillEngler (approved):
LGTM!

PR #97, opened 2024-06-13
Add ability for backend service to automatically add new users to Garden Users Globus group

### LINKED ISSUE #91: sidequest: auto-add authed users to Garden Users globus group if not already present
Globus Groups doesn't seem to support auto-join for v2 groups like the Garden Users group, which is a prerequisite for executing on our demo endpoint. One option is to repurpose the existing MDF globus group (see https://github.com/Garden-AI/garden/issues/462) for garden users too, but I'd like to explore some other options before we resort to that. 

One approach that comes to mind is to automatically add users to the globus group via the groups v2 api ([docs](https://groups.api.globus.org/redoc#tag/groups)) the first time they authenticate with the backend. 

# acceptance criteria
 - Owen no longer needs to manually accept join requests to the Garden Users group via email when someone wants to access the demo endpoint.

hholb:
Resolves: #91 

## Overview

This PR adds the ability for the backend service to automatically add authed users to the Garden Users Globus group. When a user accesses authenticated endpoints for the first time they are added to the group via the Globus Groups API.

## Changes
 - Added a field to the `Settings` object in `config.py` called `GARDEN_USER_GROUP_ID` which needs to be set in the environment to the UUID of the Garden Users group.

- Added a new module `src/auth/globus_groups.py` where the primary function `add_user_to_group` and its helpers are defined.

## Discussion

To enabe this functionality the backend service was made an administrator of the Garden Users group. This was accomplished by:
- Having the backend service authenticate with Globus, then request to join the group as a regular user
- Give the backend sercvice an administrator role in the group.

## Testing

Added new unit tests that:
- Verify a user is added to the group on their first authenticated request
- Ensure we only add the user to the group once -- this may be unnecessary if the Globus Groups API is idempotent
- Mock the necessary dependencies to isolate the functionality

### DIFF
--- a/garden-backend-service/src/models/base.py
+++ b/garden-backend-service/src/models/base.py
@@ -23,13 +23,15 @@ class Base(AsyncAttrs, DeclarativeBase):
     @classmethod
     async def get_or_create(cls: Type[T], db: AsyncSession, **kwargs: Any) -> T:
         obj = await cls.get(db, **kwargs)
+        created = False
 
         if not obj:
             obj = cls(**kwargs)
             await obj._asave(db)
             await db.refresh(obj)
+            created = True
 
-        return obj
+        return obj, created
 
     async def _asave(self, db: AsyncSession) -> None:
         db.add(self)
--- a/garden-backend-service/src/models/base.py
+++ b/garden-backend-service/src/models/base.py
@@ -21,7 +21,7 @@ class Base(AsyncAttrs, DeclarativeBase):
         return (await db.execute(q)).scalar_one_or_none()
 
     @classmethod
-    async def get_or_create(cls: Type[T], db: AsyncSession, **kwargs: Any) -> T:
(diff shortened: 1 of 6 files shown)


MaxTuecke (approved):
LGTM!

garden-backend-service/src/models/base.py
```diff
@@ -23,13 +23,15 @@ async def get(cls: Type[T], db: AsyncSession, **kwargs: Any) -> Optional[T]:
     @classmethod
     async def get_or_create(cls: Type[T], db: AsyncSession, **kwargs: Any) -> T:
```
OwenPriceSkelly:
```suggestion
    async def get_or_create(cls: Type[T], db: AsyncSession, **kwargs: Any) -> tuple[T, bool]:
```

OwenPriceSkelly (approved):
LGTM! especially happy with how you organized the new code/module; I'm sure I'm biased but everything seems like it's in "the obvious place" if we need to make tweaks in the future, we love to see it 😤

PR #98, opened 2024-06-17
Revert "Add ability for backend service to automatically add new users to Garden Users Globus group"

OwenPriceSkelly:
Reverts Garden-AI/garden-backend#97

when we deployed #97 to dev, it didn't work as intended. At least part of this is due to us not requesting the `all` scope from the globus groups client on the CLI side ([issue](https://github.com/Garden-AI/garden/issues/462)), but that's not the whole story. To check, I tried kludging my local install to use the `all` groups scope, then logged in to globus from my personal email (i.e. as someone not already part of the group or in the DB) to get a fresh bearer token, then tried hitting the dev api with 
```
curl -X 'GET' \
  'https://api-dev.thegardens.ai/greet' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <token>'
```
and still got a 500 error in response. So there's something we need to tweak in #97 before globus auth/groups is happy. 

For reference, here are the logs from lightsail: 
```
[17/Jun/2024:21:34:12] 2024-06-17 21:34:12,329 INFO sqlalchemy.engine.Engine SELECT users.id, users.username, users.identity_id
[17/Jun/2024:21:34:12] FROM users
[17/Jun/2024:21:34:12] WHERE users.username = $1::VARCHAR AND users.identity_id = $2::UUID
[17/Jun/2024:21:34:12] 2024-06-17 21:34:12,329 INFO sqlalchemy.engine.Engine [generated in 0.00017s] ('owenpriceskelly@gmail.com', UUID('e134fbc8-bbb0-4cc0-833a-42ecdbac299b'))
[17/Jun/2024:21:34:12] 2024-06-17 21:34:12,334 INFO sqlalchemy.engine.Engine INSERT INTO users (username, identity_id) VALUES ($1::VARCHAR, $2::UUID) RETURNING users.id
[17/Jun/2024:21:34:12] 2024-06-17 21:34:12,334 INFO sqlalchemy.engine.Engine [generated in 0.00018s] ('owenpriceskelly@gmail.com', UUID('e134fbc8-bbb0-4cc0-833a-42ecdbac299b'))
[17/Jun/2024:21:34:12] 2024-06-17 21:34:12,338 INFO sqlalchemy.engine.Engine SELECT users.id, users.username, users.identity_id
[17/Jun/2024:21:34:12] FROM users
[17/Jun/2024:21:34:12] WHERE users.id = $1::INTEGER
[...]

PR #99, opened 2024-06-18
Hholb/91 auto add users to garden group

### LINKED ISSUE #91: sidequest: auto-add authed users to Garden Users globus group if not already present

hholb:
Resolves: #91
Related: #97 #98 https://github.com/Garden-AI/garden/pull/486

## Overview

This PR adds the ability for the backend service to automatically add authed users to the Garden Users Globus group. When a user accesses authenticated endpoints for the first time they are added to the group via the Globus Groups API.

## Changes

- Added a field to the Settings object in config.py called GARDEN_USER_GROUP_ID which needs to be set in the environment to the UUID of the Garden Users group.

- Added a new module src/auth/globus_groups.py where the primary function add_user_to_group and its helpers are defined.

## Discussion

To enabe this functionality the backend service was made an administrator of the Garden Users group. This was accomplished by:

- Having the backend service authenticate with Globus, then request to join the group as a regular user
- Give the backend sercvice an administrator role in the group.

## Testing

Added new unit tests that:
- Verify a user is added to the group on their first authenticated request
- Mock the necessary dependencies to isolate the functionality

Manually removed myself from the group and then authenticated with the backend. I am back in the group! Adding myself more than once has no ill effects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -105,6 +105,7 @@ celerybeat.pid
 
 # Environments
 .env
+*.env
 .env.postgres
 .venv
 env/
--- a/garden-backend-service/.postgres.env
+++ b/garden-backend-service/.postgres.env
@@ -1,3 +1,0 @@
-POSTGRES_DB=garden_db_dev
-POSTGRES_USER=owenpriceskelly@uchicago.edu
-POSTGRES_PASSWORD=your_password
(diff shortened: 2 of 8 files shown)


OwenPriceSkelly (approved):
LGTM!

PR #101, opened 2024-06-21
enhancement: mock_db_session fixture

### LINKED ISSUE #94: enhancement: async postgres(-like) DB fixtures for unit tests

hholb:
Resolves: #94

## Overview

This PR sets up the test suite to run integration tests alongside a postgres instance provided by the [testcontainers](https://testcontainers-python.readthedocs.io/en/latest/) library. Tests that need to interact with the database can use the `mock_db_session` fixture defined in `conftest.py`. The `mock_db_session` fixture handles overriding the `get_db_session` dependency used by api endpoints and provides each test with a clean database to run queries on.

This PR also restructures the test sure to mirror the structure of the `src/` directory.
```text
.
└── tests
    ├── api
    │   └── routes
    │       ├── test_docker_push_token.py
    │       ├── test_doi.py
    │       ├── test_entrypoint.py
    │       ├── test_garden.py
    │       ├── test_greet.py
    │       ├── test_notebook.py
    │       └── test_routes.py
    ├── auth
    │   └── test_globus_groups.py
    ├── conftest.py
    ├── fixtures
    │   ├── EntrypointCreateRequest-shared-entrypoint.json
    │   ├── EntrypointCreateRequest-with-metadata.json
    │   ├── GardenCreateRequest-shares-entrypoint.json
    │   └── GardenCreateRequest-two-entrypoints.json
    ├── __init__.py
    └── README.md
```

## Discussion

The `mock_db_session` fixture depends on docker being available in the environment the tests are running in. To disable these tests run `pytest -m "not integration"`.  Any new tests that use the `mock_db_session` fixture should add a `@pytest.mark.integration` decorator. If docker is not present in the environment the tests that use `mock_db_session` are automatically skipped.

I had issues running the GitHub Actions workflow while testing. Steps running after the `abatilo/actions-poetry@v3` action could not find poetry on the PATH so I changed the workflow to install poetry via pipx and add poetry to PATH manually.

## Testing
Run the test suite and behold:
```bash
pytest

pytest -m "integration"

pytest -m "not integration"
```

Run the ci workflow locally with [act](https://github.com/nektos/act): `act -j build`


.github/workflows/ci.yaml
```diff
@@ -21,14 +22,26 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Install poetry
-        uses: abatilo/actions-poetry@v3
+      - name: Install pipx
```
OwenPriceSkelly, line 25: good catch on this! The `abatilo/actions-poetry@v3` claims to already install with pipx, wonder why it was breaking. Maybe something to do with `$GITHUB_PATH` variable? could be worth opening an issue for the maintainer if there isn't one already

garden-backend-service/.flake8
```diff
@@ -1,3 +1,4 @@
 [flake8]
 max-line-length = 150
-extend-ignore = E203
+extend-ignore = E203, PT004
+exclude = migrations,__pycache__,schemas
```
OwenPriceSkelly: why ignore schemas? (assuming this matches `src/api/schemas`)

garden-backend-service/src/api/routes/entrypoint.py
```diff
@@ -22,6 +22,7 @@ async def add_entrypoint(
     db.add(new_entrypoint)
     try:
         await db.commit()
+        await db.refresh(new_entrypoint)
```
OwenPriceSkelly: this seems reasonable - was this necessary to fix a bug with the tests or just better practice?

garden-backend-service/tests/README.md
OwenPriceSkelly, line 1: love the readme!!

garden-backend-service/tests/conftest.py
```diff
@@ -0,0 +1,209 @@
...
+    _get_auth_token,
+)
+
+from src.config import Settings, get_settings
+from src.main import app
+from src.models.base import Base
+
+
+@pytest.fixture
+def client(patch_globus_groups):
```
OwenPriceSkelly, line 25: understanding check: what's the difference between this client and the previous `from fastapi.testclient import TestClient`?

garden-backend-service/tests/conftest.py
```diff
@@ -0,0 +1,209 @@
...
+    mocker.patch("src.api.dependencies.auth.add_user_to_group")
+
+
+def docker_available():
+    available = shutil.which("docker")
+    return available
+
+
+def pytest_collection_modifyitems(session, config, items):
+    """Skip integration tests that rely on docker if docker is not availablie"""
```
OwenPriceSkelly:
```suggestion
    """Skip integration tests that rely on docker if docker is not available"""
```

garden-backend-service/tests/conftest.py
```diff
@@ -0,0 +1,209 @@
...
+def patch_globus_groups(mocker):
+    mocker.patch("src.api.dependencies.auth.add_user_to_group")
+
+
+def docker_available():
+    available = shutil.which("docker")
+    return available
+
+
+def pytest_collection_modifyitems(session, config, items):
```
OwenPriceSkelly, line 40: This is very slick, big fan

OwenPriceSkelly (approved):
LGTM! This has me excited to write tests, which is a _very_ new feeling for me. Couple of questions just for my own understanding but nothing blocking - merge away 🫡

garden-backend-service/.flake8
hholb: Oh, whoops! I added this to my local config while working on this to suppress some flake8 complaints while I worked on the tests. I meant to take it out before pushing.

MaxTuecke (approved):
Very thorough tests, this looked like a fun one to figure out. As far as i can tell, it LGTM!

garden-backend-service/src/api/routes/entrypoint.py
hholb: I thought some of the entrypoint tests were failing due to the relationships not being updated when adding a new one to the database. Turns out that wasn't the issue and the call to refresh is unnecessary. The tests pass without it.

garden-backend-service/tests/conftest.py
hholb: The `fastiapi.TestClient` is built on the `httpx.Client` which only supports synchronous requests. The fastapi [docs](https://fastapi.tiangolo.com/advanced/async-tests/) recommend using the `httpx.AsyncClient` in async tests.

garden-backend-service/tests/conftest.py
hholb: Good catch!

PR #103, opened 2024-06-21
Ownership / restricted deletes

### LINKED ISSUE #100: revamp: basic access control / restricted deletes
As a user who has published work with garden, I don't want anyone but me to be able to delete or overwrite my work (intentionally or not). 

As the responsible steward of a DOI, I should only be able to delete an entrypoint or garden if its DOI is still in draft state. 

## Acceptance criteria 
  - gardens/entrypoints have references to an "owner" user  
  - If a DELETE request is sent by a non-owner and/or for a non-draft DOI, it fails with a helpful explanation

OwenPriceSkelly:
closes #100 
## Overview
This implements some minimal access control for gardens and entrypoints by linking them to an "owner" user id. Deletes are now prevented in the following scenarios: 
  - the DOI is registered 
  - the user making the DELETE request is not the owner 
  - (prior PR) there is still a garden referring to the entrypoint being deleted

## Discussion

Because we hadn't been including the `is_doi_draft` field for gardens on the search index, now whenever you POST a new garden without that field, we'll check with doi.org to set it to the correct value. 

By default, the owner is the `authed_user` behind the POST request to create the garden or entrypoint, but a globus auth identity_id may also be explicitly specified to create under a different owner. If _that_ user doesn't exist in the DB, we fall back to the `authed_user` making the request and log a warning. I'm not sure that we'll have much use for this once I have the search index migration complete, but it seemed nice to have the option and made testing easier. 

I also had to modify the alembic migration script to handle existing gardens/entrypoints without owners, since I wanted it to be non-nullable. I made myself the default owner in those cases: all your tutorial garden are belong to me 😈. 

## Testing

Just manual testing so far, but I've been throwing the whole dev search index at these routes locally. Unit tests are either coming in this PR or a follow-up PR depending on whether we get @hholb's #101 in ahead of it or not.


WillEngler (approved):
LGTM! This is a really nice milestone - we've been talking about lack of access control for a long time and you finally got us here

Issue #52, opened 2023-12-22
Infra: Add tags to all Garden resources


WillEngler:
We'd like to track costs for Garden vs the other Accelerate services. To do this consistently, we should add tags in our Terraform. And then we can track spending based on the tags.

PR #108, opened 2024-06-28
Add EC2 instance to connect pgadmin with RDS instances

### LINKED ISSUE #102: infra: make it possible to connect pgadmin to rds instances
our RDS instances aren't publicly available, so in order to connect to them via something like pgadmin we need an ec2 instance in our default VPC we can use to set up an ssh tunnel. 

this probably isn't very much work to set up, just making an issue for it before I forget

hholb:
Resolves: #102 

Starts resolving: #52 -- added `Project = "Garden"` to resource tags in the `rds` module.

## Overview

This PR adds resources to the `rds` Terraform module that provision an EC2 instance that can communicate with the RDS instance and act as an ssh tunnel allowing a local pgadmin to connect with the RDS instance.

## Discussion

This setup requires ssh keys to be added to the EC2 instance manually. I have included instructions on how to do that and how to configure pgadmin in `README.md`.

## Testing
Tested manually in my personal AWS account using a similar setup.

`terraform plan` looks correct against the current dev terraform state.

### DIFF
--- a/garden-backend-service/README.md
+++ b/garden-backend-service/README.md
@@ -117,43 +117,48 @@ On a push to `dev` or `prod` branches, we run a GitHub action to build and push 
 
 ## Connect pgadmin to RDS instances
 
-Note: You will need to follow this process for each instance. You can
+**Note:** You will need to follow this process for each instance (dev and prod). You can
 use the same key pair for both instances.
 
-1. Create ssh key pair
+### **Create ssh key pair**
+   
 Either through the AWS console or on the command line create an RSA
 keypair.
 
-- Option 1 AWS Console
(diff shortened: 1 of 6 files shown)


garden-backend-service/README.md
OwenPriceSkelly, line 1: really excellent documentation of the process, exactly the right amount of detail IMO

OwenPriceSkelly (approved):
LGTM! thanks for pulling this together, plus great documentation - following your readme I was able to connect on the first try 🙌 (though I did make some tweaks where the default username turned out to be `ubuntu` instead of `ec2-user`). 

Good stuff!

PR #110, opened 2024-07-02
add PUT routes for gardens/entrypoints

### LINKED ISSUE #106: feat: enable updates to gardens/entrypoints
Can't spell CRUD without U! 

The first passes at the /garden and /entrypoint resources only implemented operations for creating, reading and deleting entire gardens or entrypoints, but we'll need the ability to update/replace existing resources too. 

## Acceptance criteria: 
  - PUT /garden/{doi} and PUT /entrypoint/{doi} routes implemented 
  - These should be basically equivalent to respective POST methods when the resource doesn't exist yet
  - If the resource identified by {doi} exists, it should be updated with the data provided in the request body. The request body should completely describe a resource, whether it's to create a new one or update an existing one
  - Should not allow users to modify existing resources that do not belong to them 
  - repeated PUTs should be idempotent

OwenPriceSkelly:
closes #106 

## Overview

As of these last two routes, we officially have "`local_data` parity" for the backend 🎉 🥳 

calling `PUT /entrypoints/{doi}` (or `/gardens/{doi}`) for an entrypoint that does not yet exist is exactly the same behavior as `POST /entrypoints` would with the same request body. 

For data that does already exist, updating requires the same ownership permissions that deleting and recreating the resource would. 

## Discussion

This supports updates in a way that is deliberately similar to local data, where an entry is updated by overwriting it with a complete replacement. In the future we might want to implement a PATCH method for partial updates. 

This also demotes the 'related metadata' tables to mere `ARRAY(JSON)` fields on entrypoints. This is more flexible until we've got the migration finished, and once we have a more concrete need for e.g. a `dataset_metadata` table we can bring it back. 

## Testing

Sanity checked locally with the hacky migration script I've been building on the side. Also added some happy-path tests for both gardens and entrypoints to test creating and updating with the new routes. 

Specifically also made sure that updating the many-many relationship works as expected -- you can drop a previously connected entrypoint from a garden by simply omitting the DOI from `entrypoint_ids` in your update (like we trivially can with local_data)


garden-backend-service/src/api/routes/entrypoints.py
```diff
@@ -98,3 +71,75 @@ async def delete_entrypoint(
...
+        explicit_owner: User | None = await User.get(
+            db, identity_id=entrypoint.owner_identity_id
+        )
+        if explicit_owner is not None:
+            owner = explicit_owner
+        else:
+            logger.warning(
+                f"No user found with Globus identity ID {entrypoint.owner_identity_id}. "
+                f"Assigning default ownership to {user.identity_id} ({user.username}). "
+            )
```
WillEngler, line 130: Is the no-owner logic something we could get rid of after the migration? (Not indicating that this is bad, just curious)

garden-backend-service/src/api/routes/gardens.py
```diff
@@ -109,3 +62,108 @@ async def delete_garden(
...
+    db: AsyncSession,
+    user: User,
+):
+    # if not specified, check draft status with the real world (doi.org)
+    if garden_data.doi_is_draft is None:
+        registered = await is_doi_registered(garden_data.doi)
+        garden_data.doi_is_draft = not registered
+
+    # collect entrypoints by DOI
+    entrypoints = await _collect_entrypoints(garden_data.entrypoint_ids, db)
```
WillEngler, line 138: I'm curious, why do we need the fully hydrated Entrypoint objects to create a Garden? It makes sense that we need to check if the entrypoint ids the user provided correspond to entrypoints in the DB. But why do we need to stick the full objects on to `new_garden`?

WillEngler: Oh I might have a guess ... in the 201 CREATED response you want to send back the whole garden including its full entrypoint metadata in the response. So that's why it's useful to grab the full Entrypoint objects. (maybe?)

garden-backend-service/src/models/entrypoint.py
```diff
@@ -42,33 +32,17 @@ class Entrypoint(Base):
...
+    tags: Mapped[list[str]] = mapped_column(ARRAY(String))
+    test_functions: Mapped[list[str]] = mapped_column(ARRAY(String))
+    requirements: Mapped[list[str] | None] = mapped_column(ARRAY(String))
+
+    # NOTE: modifications to these lists / dictionaries won't be picked up
+    # by sqlalchemy ORM. Updates should replace with a copy.
+    models: Mapped[list[dict] | None] = mapped_column(ARRAY(JSON))
+    repositories: Mapped[list[dict] | None] = mapped_column(ARRAY(JSON))
+    papers: Mapped[list[dict] | None] = mapped_column(ARRAY(JSON))
+    datasets: Mapped[list[dict] | None] = mapped_column(ARRAY(JSON))
```
WillEngler, line 45: 🫡 farewell for now full metadata tables. Hello to our old friend JSON

WillEngler (approved):
This looks good to me! Some things I'm curious about but nothing I see that needs changing.

garden-backend-service/src/api/routes/entrypoints.py
OwenPriceSkelly: yeah that's more or less what I'm thinking, seemed like a reasonably convenient middle ground until we need a more sophisticated ownership model than "only one user can touch this"

garden-backend-service/src/api/routes/gardens.py
OwenPriceSkelly: assigning the full entrypoint objects to the garden was honestly just the first thing I attempted that got sqlalchemy to update the many-many association table correctly. definitely interested if I was missing a better way to do that with the ORM

PR #113, opened 2024-07-03
Update users table to support frontend profile page

### LINKED ISSUE #105: users table support for profile info
Currently, all we persist about users is their globus auth identity_id and their username (typically their email), both of which are inferred from their auth token. Now that we have @jamies31 working on the user profile page for the frontend, we'll have to define some additional fields as well as some new `/user` routes to support that work. 

## New fields
The info we want the ability to display in the first pass is: 
  - [ ] first / last names 
  - [x] email address 
  - [ ] phone number
  - [ ] institutional affiliations
  - [ ] skills 
  - [ ] domains
  - [ ] saved gardens 

where possible (e.g. email address, possibly name and affiliations) we should infer this from the user's globus ID, but we should still allow users to edit that info in case it changes or the inferred data from globus auth is not correct. The authed user should only be able to edit their own info. 

## Other thoughts / anticipating scope creep: 
  - for skills/domains, kj and jamie were thinking it would be nice to have a few options for users to select from rather than manually entering them. Since we're almost definitely want to also allow users to type them in manually, we'll probably want to leave that auto-fill feature for another story and treat skills/domains like tags for now. 
  - likewise, the ability to save gardens would be its own story, but we can at least define the relationship for the `saved_gardens` field here.

hholb:
Resolves: #105 

## Overview

This PR adds the following fields to users table:
- name
- email
- phone_number
- skills
- domains
- affiliations
- saved_gardens
- profile_pic_id (at the request of @jamies31)

To track saved_gardens, a new relation called `users_saved_gardens` is added to the database.

This PR adds a `/users` endpoint that supports GET and PATCH requests to fetch and update data for authenticated users. 

## Testing

Added new unit tests that exercise the new endpoints.


garden-backend-service/src/api/routes/users.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+@router.patch("", status_code=status.HTTP_200_OK, response_model=UserMetadataResponse)
+async def update_user_info(
+    user_update: UserUpdateRequest,
+    db: AsyncSession = Depends(get_db_session),
+    authed_user: User = Depends(authed_user),
+):
+    """Update the user profile info for the current authed user."""
+    update_data: Dict[str, Any] = user_update.dict(exclude_unset=True)
```
OwenPriceSkelly: style nits for a python 3.11 / pydantic 2.x world 
```suggestion
    update_data: dict[str, Any] = user_update.model_dump(exclude_unset=True)
```

garden-backend-service/src/api/schemas/user.py
```diff
@@ -0,0 +1,24 @@
+from uuid import UUID
+
+from pydantic import BaseModel, EmailStr
+
+from .base import UniqueList
+
+
+class UserBase(BaseModel):
```
OwenPriceSkelly: let's use our `BaseSchema` (also from `.base`) instead of `BaseModel` so that our schemas share config. That way, when fastAPI converts an ORM instance into a response json, it will behave the same way for all of our response schemas.

garden-backend-service/src/api/routes/users.py
```diff
@@ -0,0 +1,37 @@
...
+@router.get("", status_code=status.HTTP_200_OK, response_model=UserMetadataResponse)
+async def get_user_info(
+    authed_user: User = Depends(authed_user),
+):
+    """Return the user profile info for the current authed user."""
+    return authed_user
+
+
+@router.patch("", status_code=status.HTTP_200_OK, response_model=UserMetadataResponse)
+async def update_user_info(
```
OwenPriceSkelly, line 40: this looks good, but I'm noticing that we can't use this route yet to actually update the user's saved gardens. Totally reasonable call if you think that's out of scope for this PR (especially since we may want updates to a relationship to be its own route like `/users/{id}/saved_gardens`), just wanted to check that it was intentional so I can make a separate issue for it

garden-backend-service/src/api/schemas/user.py
```diff
@@ -0,0 +1,24 @@
...
+    username: str | None = None
+    name: str | None = None
+    email: EmailStr | None = None
+    phone_number: str | None = None
```
OwenPriceSkelly: might be nice to validate this with the `pydantic-extra-types` phone number type

garden-backend-service/src/models/_associations.py
```diff
@@ -7,3 +7,10 @@
     Column("garden_id", ForeignKey("gardens.id"), primary_key=True),
     Column("entrypoint_id", ForeignKey("entrypoints.id"), primary_key=True),
 )
+
+users_saved_gardens = Table(
+    "users_saved_gardens",
+    Base.metadata,
+    Column("user_id", Integer, ForeignKey("users.id")),
```
OwenPriceSkelly: I think we also want the `primary_key=True` constraint for these columns (like the gardens_entrypoints table)

garden-backend-service/src/models/garden.py
```diff
@@ -41,3 +41,7 @@ class Garden(Base):
     user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
     user: Mapped[User] = relationship(lazy="selectin")
     owner: Mapped[User] = synonym("user")
+
+    users: Mapped[list[User]] = relationship(
```
OwenPriceSkelly: I don't think we want this attribute on the garden side if we don't need it. User entities definitely want easy access to their saved gardens but I think we should prefer an explicit query if we need to support "all users who have saved this garden" in the future

OwenPriceSkelly: (i.e. the same many-many pattern as gardens to entrypoints)

garden-backend-service/src/models/user.py
```diff
@@ -1,11 +1,35 @@
+from typing import TYPE_CHECKING, Optional
```
OwenPriceSkelly: style nit: let's use `... | None` instead of `Optional[...]` syntax for type annotations. partly on aesthetic grounds, but I also have a gripe with the way that "optional" can mean "has a default" or "nullable" depending on context, while `| None` is unambiguous

garden-backend-service/src/models/user.py
```diff
@@ -1,11 +1,35 @@
...
+
 
 class User(Base):
     __tablename__ = "users"
     id: Mapped[int] = mapped_column(primary_key=True)
-    username: Mapped[str]
+    username: Mapped[Optional[str]]
     identity_id: Mapped[UUID] = mapped_column(unique=True)
+    name: Mapped[Optional[str]]
+    email: Mapped[Optional[str]] = mapped_column(unique=True)
```
OwenPriceSkelly: I don't think email should be unique, especially if we're allowing users to fill it themselves. for example, several users might want the email displayed on their profile to be "labs@globus.com" or something like that

garden-backend-service/src/models/user.py
```diff
@@ -1,11 +1,35 @@
...
+    phone_number: Mapped[Optional[str]]
+    skills: Mapped[Optional[list[str]]] = mapped_column(postgresql.ARRAY(String))
+    domains: Mapped[Optional[list[str]]] = mapped_column(postgresql.ARRAY(String))
+    affiliations: Mapped[Optional[list[str]]] = mapped_column(postgresql.ARRAY(String))
+
+    saved_gardens: Mapped[list["Garden"]] = relationship(
+        "Garden",
+        secondary=users_saved_gardens,
+        back_populates="users",
+        lazy="selectin",
```
OwenPriceSkelly: like we talked about when we synced up earlier, I don't think we should eagerly load the saved_gardens if we don't have to. Lazy-loading will make things trickier in an async context, but since gardens also eagerly load entrypoints we're liable to load way more nested data than we need just to build a list of dois

garden-backend-service/src/models/user.py
```diff
@@ -1,11 +1,35 @@
...
+    )
+
+    @property
```
OwenPriceSkelly: I won't pretend to have any experience with using descriptors in sqlalchemy models, but it looks like plain `@property` is not usually the preferred option: https://docs.sqlalchemy.org/en/20/orm/mapped_sql_expr.html

OwenPriceSkelly: on second thought, something like the "last resort" [example](https://docs.sqlalchemy.org/en/20/orm/mapped_sql_expr.html#using-a-plain-descriptor)  using `object_session` to select just the dois might be what we want here. Either that or just building the list of dois in the route itself, which might be simpler to work around with async

OwenPriceSkelly (changes requested):
This is a great first pass, the only substantial change that might take some special care is how we should be loading the related saved_gardens (like we talked about yesterday), otherwise this is in really good shape. excellent tests, as usual 🙌

garden-backend-service/src/api/routes/users.py
hholb: I was thinking that would be a separate route, but it wouldn't be an issue to add that functionality here if that is what we want. I am no expert on API design, so I am happy to go either way.

OwenPriceSkelly: let's leave it off for now and pull it into its own issue we can come back to either way

hholb: #116

garden-backend-service/migrations/versions/f293268b31c5_add_fields_to_users_table_create_users_.py
```diff
@@ -0,0 +1,65 @@
...
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "users_saved_gardens",
+        sa.Column("user_id", sa.Integer(), nullable=True),
+        sa.Column("garden_id", sa.Integer(), nullable=True),
```
OwenPriceSkelly: I think the change to add the primary key constraint slipped through. IMO nothing wrong with having multiple migration scripts in a single PR in the future so we can keep using --autogenerate instead of doing stuff by hand 
```suggestion
        sa.Column("user_id", sa.Integer(), nullable=False),
        sa.Column("garden_id", sa.Integer(), nullable=False),
```

garden-backend-service/src/api/routes/users.py
```diff
@@ -0,0 +1,62 @@
...
+from src.api.dependencies.database import get_db_session
+from src.api.schemas.user import UserMetadataResponse, UserUpdateRequest
+from src.models import User
+from src.models._associations import users_saved_gardens
+
+router = APIRouter(prefix="/users")
+
+
+async def _get_saved_garden_dois(user: User, db: AsyncSession) -> list[str]:
+    from src.models import Garden  # import here to avoid circular imports
```
OwenPriceSkelly: This surprised me (since we `from src.models import User` just fine above) so I tried moving it up and I'm not running into a circular import error. can you still reproduce it on your end?

garden-backend-service/src/api/routes/users.py
```diff
@@ -0,0 +1,62 @@
+from typing import Any
+
+from fastapi import APIRouter, Depends, status
+from sqlalchemy import select
+from sqlalchemy.ext.asyncio import AsyncSession
+from src.api.dependencies.auth import authed_user
+from src.api.dependencies.database import get_db_session
+from src.api.schemas.user import UserMetadataResponse, UserUpdateRequest
+from src.models import User
```
OwenPriceSkelly:
```suggestion
from src.models import User, Garden
```

garden-backend-service/src/api/routes/users.py
```diff
@@ -0,0 +1,62 @@
...
+from src.models._associations import users_saved_gardens
+
+router = APIRouter(prefix="/users")
+
+
+async def _get_saved_garden_dois(user: User, db: AsyncSession) -> list[str]:
+    from src.models import Garden  # import here to avoid circular imports
+
+    result = await db.execute(
```
OwenPriceSkelly, line 16: I think it's exactly equivalent to what you have but FYI the `.scalars` method instead of `execute` could skip the row objects

garden-backend-service/tests/conftest.py
```diff
@@ -111,6 +111,8 @@ def mock_auth_state():
     mock_auth.username = "Monsieur.Sartre@ens-paris.fr"
     mock_auth.identity_id = UUID("00000000-0000-0000-0000-000000000000")
     mock_auth.token = "tokentokentoken"
+    mock_auth.email = "some@email.com"
+    mock_auth.name = "M. Sartre"
```
OwenPriceSkelly, line 115: 🥖🍷

OwenPriceSkelly (approved):
LGTM!

Issue #112, opened 2024-07-03
migration: deprecate `/garden-search-record` routes


OwenPriceSkelly:
After we complete the migration, we may still have users with older versions of the garden sdk that try to hit the `/garden-search-record` routes to publish/delete their gardens. Rather than just deleting the routes entirely, those users should be given a deprecation message instructing them to update their version of the sdk.

PR #115, opened 2024-07-09
Deprecate '/garden-search-record' routes, keep search index in sync with db

### LINKED ISSUE #95: revamp: ensure search index is in sync with database
Assumptions: 
  - we've migrated the contents of the search index to the DB 
  - we've stopped writing to directly (and deprecated the `/garden-search-record` route)
  

We'll want to keep the search index handy, but it should be kept in sync with the database contents and not the other way around. 

Acceptance criteria: 
  - The database is the One Source Of Truth.

hholb:
Resolves: #95, #112 

## Overview

This PR deprecates the `/garden-search-record` routes and adds logic to keep the search index in sync with our database.

## Discussion

I created a new `@deprecated` decorator we can add to routes that returns 410 status code and a helpful message to clients with an optional link to relevant docs. For example:
```
@router.post("", status_code=status.HTTP_200_OK, deprecated=True)
@deprecated(
    name="/garden-search-record",
    message="Use /gardens instead or update garden-ai package to latest version.",
    doc_url="https://api-dev.thegardens.ai/docs#/default/add_garden_gardens_post",
)
async def publish_search_record(
    garden_meta: PublishSearchRecordRequest,
    settings: Settings = Depends(get_settings),
    search_client: SearchClient = Depends(get_globus_search_client),
    _auth: AuthenticationState = Depends(authenticated),
):
```
This allows us to deprecate routes without changing the function body.

Synchronizing the search index with our database is accomplished using [background tasks](https://fastapi.tiangolo.com/tutorial/background-tasks/). In endpoints that create, update, or delete garden data, if the updates to our database succeed, a background task is scheduled to update the search index. This should be transparent to clients.

Synchronizing with the search index can be toggled on and off using a new config variable called `SYNC_SEARCH_INDEX`. It is set to False in the test suite so we don't accidentally send test data to the search index.

## Testing

I have not found a good way to test the synchronization with the search index. My first thought was to patch the functions that update the search index and in the tests for garden routes do something like:
```
mock_update_search_index = mocker.patch(...)
# Run the rest of the test....
mock_update_search_index.assert_called_once()
```
But, because background tasks happen after the endpoint function returns this assertion is flaky.

### DIFF
--- a/garden-backend-service/src/api/routes/_tasks.py
+++ b/garden-backend-service/src/api/routes/_tasks.py
@@ -1,25 +1,32 @@
 from src.api.dependencies.search import get_globus_search_client
 from src.auth.globus_auth import get_auth_client
 from src.config import Settings
+from src.models import Garden
 
-from ._utils import poll_globus_search_task
+from ._utils import _garden_sqlalchemy_to_pydantic, poll_globus_search_task
 
 
-async def delete_from_search_index(garden_data, settings: Settings):
+async def delete_from_search_index(garden: Garden, settings: Settings):
     client = get_globus_search_client(get_auth_client())
     delete_result = client.delete_entry(
         settings.GLOBUS_SEARCH_INDEX_ID,
-        garden_data.doi,
+        garden.doi,
     )
     task_id = delete_result["task_id"]
     return await poll_globus_search_task(task_id, client)
 
 
-async def create_or_update_on_search_index(garden_data, settings: Settings):
+async def create_or_update_on_search_index(garden: Garden, settings: Settings):
     client = get_globus_search_client(get_auth_client())
+    garden_pub = _garden_sqlalchemy_to_pydantic(garden)
+    garden_meta = {
+        "subject": garden_pub.doi,
+        "visible_to": ["public"],
+        "content": garden_pub.dict(),
+    }
     create_result = client.create_entry(
         settings.GLOBUS_SEARCH_INDEX_ID,
-        garden_data.doi,
+        garden_meta,
     )
     task_id = create_result["task_id"]
 
--- a/garden-backend-service/src/api/routes/_tasks.py
+++ b/garden-backend-service/src/api/routes/_tasks.py
@@ -1,0 +1,28 @@
+from src.api.dependencies.search import get_globus_search_client
+from src.auth.globus_auth import get_auth_client
+from src.config import Settings
+
+from ._utils import poll_globus_search_task
+
+
+async def delete_from_search_index(garden_data, settings: Settings):
+    if settings.SYNC_SEARCH_INDEX:
+        client = get_globus_search_client(get_auth_client())
+        delete_result = client.delete_entry(
+            settings.GLOBUS_SEARCH_INDEX_ID,
+            garden_data.doi,
+        )
+        task_id = delete_result["task_id"]
+        return await poll_globus_search_task(task_id, client)
+
+
+async def create_or_update_on_search_index(garden_data, settings: Settings):
+    if settings.SYNC_SEARCH_INDEX:
+        client = get_globus_search_client(get_auth_client())
+        create_result = client.create_entry(
+            settings.GLOBUS_SEARCH_INDEX_ID,
+            garden_data.doi,
+        )
+        task_id = create_result["task_id"]
+
+        return await poll_globus_search_task(task_id, client)
--- a/garden-backend-service/src/api/routes/_tasks.py
+++ b/garden-backend-service/src/api/routes/_tasks.py
@@ -6,23 +6,21 @@ from ._utils import poll_globus_search_task
 
 
 async def delete_from_search_index(garden_data, settings: Settings):
-    if settings.SYNC_SEARCH_INDEX:
-        client = get_globus_search_client(get_auth_client())
-        delete_result = client.delete_entry(
-            settings.GLOBUS_SEARCH_INDEX_ID,
-            garden_data.doi,
-        )
-        task_id = delete_result["task_id"]
-        return await poll_globus_search_task(task_id, client)
+    client = get_globus_search_client(get_auth_client())
+    delete_result = client.delete_entry(
+        settings.GLOBUS_SEARCH_INDEX_ID,
+        garden_data.doi,
+    )
+    task_id = delete_result["task_id"]
+    return await poll_globus_search_task(task_id, client)
 
 
 async def create_or_update_on_search_index(garden_data, settings: Settings):
-    if settings.SYNC_SEARCH_INDEX:
-        client = get_globus_search_client(get_auth_client())
-        create_result = client.create_entry(
-            settings.GLOBUS_SEARCH_INDEX_ID,
-            garden_data.doi,
-        )
-        task_id = create_result["task_id"]
+    client = get_globus_search_client(get_auth_client())
+    create_result = client.create_entry(
+        settings.GLOBUS_SEARCH_INDEX_ID,
+        garden_data.doi,
+    )
+    task_id = create_result["task_id"]
 
-        return await poll_globus_search_task(task_id, client)
+    return await poll_globus_search_task(task_id, client)
--- a/garden-backend-service/src/api/routes/_tasks.py
+++ b/garden-backend-service/src/api/routes/_tasks.py
@@ -1,9 +1,10 @@
 from src.api.dependencies.search import get_globus_search_client
+from src.api.schemas.search import PublishSearchRecordRequest
 from src.auth.globus_auth import get_auth_client
 from src.config import Settings
 from src.models import Garden
 
-from ._utils import _garden_sqlalchemy_to_pydantic, poll_globus_search_task
+from ._utils import poll_globus_search_task
 
 
 async def delete_from_search_index(garden: Garden, settings: Settings):
@@ -18,11 +19,11 @@ async def delete_from_search_index(garden: Garden, settings: Settings):
 
 async def create_or_update_on_search_index(garden: Garden, settings: Settings):
     client = get_globus_search_client(get_auth_client())
-    garden_pub = _garden_sqlalchemy_to_pydantic(garden)
+    garden_pub = PublishSearchRecordRequest.model_validate(garden, from_attributes=True)
     garden_meta = {
         "subject": garden_pub.doi,
         "visible_to": ["public"],
-        "content": garden_pub.dict(),
+        "content": garden_pub.model_dump(mode="json"),
     }
     create_result = client.create_entry(
         settings.GLOBUS_SEARCH_INDEX_ID,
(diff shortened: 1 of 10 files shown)


garden-backend-service/src/api/routes/_utils.py
```diff
@@ -44,3 +50,73 @@ async def is_doi_registered(doi: str) -> bool:
         return True
     else:
         return False
+
+
+def deprecated(
```
OwenPriceSkelly, line 67: I'm such a sucker for decorators, big fan

garden-backend-service/src/api/routes/_utils.py
```diff
@@ -44,3 +50,73 @@ async def is_doi_registered(doi: str) -> bool:
...
+    name="Endpoint",
+    message: str = None,
+    doc_url: str = "https://api-dev.thegardens.ai/docs",
```
OwenPriceSkelly: lets point this at the prod api endpoint by default 
```suggestion
    doc_url: str = "https://api.thegardens.ai/docs",
```

garden-backend-service/src/api/routes/_utils.py
```diff
@@ -44,3 +50,73 @@ async def is_doi_registered(doi: str) -> bool:
...
+
+    if task_result["state"] == "SUCCESS":
+        return {}
+    else:
+        raise exceptions.HTTPException(
+            status.HTTP_500_INTERNAL_SERVER_ERROR, detail=task_result.text
+        )
+
+
+async def delete_from_search_index(garden_data, settings):
```
OwenPriceSkelly: organization wise, since these helpers are (unlike other utils) designed to be used specifically with BackgroundTasks, I think we could move them to their own `_tasks.py` module

garden-backend-service/src/api/routes/_utils.py
```diff
@@ -44,3 +50,73 @@ async def is_doi_registered(doi: str) -> bool:
...
+    if settings.SYNC_SEARCH_INDEX:
```
OwenPriceSkelly: I think this a bit surprising for the function `delete_from_search_index` to depend on a `SYNC_SEARCH_INDEX` setting. I think the check should happen at the call site, i.e. routes themselves should check if they should SYNC_SEARCH_INDEX, and if so send off the appropriate background task

garden-backend-service/src/api/routes/garden_search_record.py
```diff
@@ -1,17 +1,20 @@
...
 
 router = APIRouter(prefix="/garden-search-record")
 
 
-@router.post("", status_code=status.HTTP_200_OK)
+@router.post("", status_code=status.HTTP_200_OK, deprecated=True)
+@deprecated(
+    name="/garden-search-record",
+    message="Use /gardens instead or update garden-ai package to latest version.",
+    doc_url="https://api-dev.thegardens.ai/docs#/default/add_garden_gardens_post",
```
OwenPriceSkelly: ditto for the doc_url here, let's point it at prod api docs

garden-backend-service/src/api/routes/garden_search_record.py
```diff
@@ -1,17 +1,20 @@
...
-from src.api.routes._utils import is_doi_registered
+from src.api.routes._utils import deprecated, is_doi_registered, poll_globus_search_task
 from src.api.schemas.search import DeleteSearchRecordRequest, PublishSearchRecordRequest
 from src.config import Settings, get_settings
 
 router = APIRouter(prefix="/garden-search-record")
 
 
-@router.post("", status_code=status.HTTP_200_OK)
+@router.post("", status_code=status.HTTP_200_OK, deprecated=True)
```
OwenPriceSkelly, line 12: I don't think `deprecated=True` is what we want here, since we don't actually want these routes to be deprecated until we flip the switch on SYNC_SEARCH_INDEX. That way, if we discover a bug and need to un-deprecate the routes we can just change the environment variable instead of needing to change the code and redeploy.

OwenPriceSkelly: thinking through this a little I think the easiest way to do this is just check the flag in the body of the route to raise the deprecation error. but if there's a way to capture that with your decorator that could be cool too. I'd bet that naively changing this to `deprecated=get_settings().SYNC_SEARCH_INDEX` would cause problems with tests/ mocked settings but I haven't checked.

garden-backend-service/src/api/routes/gardens.py
```diff
@@ -67,13 +78,17 @@ async def delete_garden(
...
+    background_tasks: BackgroundTasks,
     garden_data: GardenCreateRequest,
     db: AsyncSession = Depends(get_db_session),
     user: User = Depends(authed_user),
+    settings: Settings = Depends(get_settings),
 ):
     existing_garden: Garden | None = await Garden.get(db, doi=doi)
     if existing_garden is None:
-        return await _create_new_garden(garden_data, db, user)
+        return await _create_new_garden(
```
OwenPriceSkelly: this smells a little off to me, I think that `_create_new_garden` shouldn't also be responsible for sending off background tasks. imo we want to dispatch background tasks from the routes themselves rather than from inside of utility functions

garden-backend-service/src/api/routes/gardens.py
```diff
@@ -99,6 +114,9 @@ async def create_or_replace_garden(
         setattr(existing_garden, key, value)
     try:
         await db.commit()
+        background_tasks.add_task(
```
OwenPriceSkelly: yeah this seems like the natural place to add the task compared to the _create_new_garden call above. I think this is also the "seam" where we want to be checking the SYNC_SEARCH_INDEX flag

OwenPriceSkelly (changes requested):
I think this is a nice proof of concept for using background tasks instead of naively writing through to the search index everywhere, and I really like the approach in general. 

Most of my feedback I had here is around where we check settings for the flag and where we actually dispatch the background tasks, but is ultimately pretty minor.  

More importantly, we need to handle updates to entrypoints -- the search index entries are gardens at the top level, but they include all the denormalized metatadata for the garden's entrypoints as well. So whenever entrypoints are edited/deleted, we'll need background tasks to update _all_ the affected gardens, too, or the search index won't be an accurate reflection of the db.

garden-backend-service/src/api/routes/_tasks.py
```diff
@@ -1,25 +1,32 @@
...
+        garden.doi,
     )
     task_id = delete_result["task_id"]
     return await poll_globus_search_task(task_id, client)
 
 
-async def create_or_update_on_search_index(garden_data, settings: Settings):
+async def create_or_update_on_search_index(garden: Garden, settings: Settings):
     client = get_globus_search_client(get_auth_client())
+    garden_pub = _garden_sqlalchemy_to_pydantic(garden)
```
OwenPriceSkelly: I think that `PublishSearchRecordRequest.model_validate(garden, from_attributes=True)` does exactly the same thing as your helper here. Test it to make sure, but if it does we should prefer the pydantic builtin over rolling our own helpers

OwenPriceSkelly: Also totally reasonable to reach for that schema, but I think we actually want to use the same `GardenMetadataResponse` schema we're using for the /gardens routes for consistency. That way we don't need to worry about the content visible on the search index diverging from content returned by the API

garden-backend-service/src/api/routes/_tasks.py
```diff
@@ -1,25 +1,32 @@
...
+    garden_meta = {
+        "subject": garden_pub.doi,
+        "visible_to": ["public"],
+        "content": garden_pub.dict(),
```
OwenPriceSkelly:
```suggestion
        "content": garden_pub.model_dump(mode='json'),
```
tbh I don't love the pydantic v2 rename `dict` -> `model_dump` either, but we might as well go along with the library's preferred convention for consistency

garden-backend-service/src/api/routes/_utils.py
```diff
@@ -110,3 +113,26 @@ async def poll_globus_search_task(
         raise exceptions.HTTPException(
             status.HTTP_500_INTERNAL_SERVER_ERROR, detail=task_result.text
         )
+
+
+def _garden_sqlalchemy_to_pydantic(garden: Garden):
```
OwenPriceSkelly: see above -- I don't think we need this if `model_validate` works instead. IIRC `model_validate` is what fastAPI calls under the hood when your route returns a plain python object to parse it into the response body

garden-backend-service/src/api/routes/_utils.py
```diff
@@ -44,3 +54,85 @@ async def is_doi_registered(doi: str) -> bool:
...
+
+async def get_gardens_for_entrypoint(
+    entrypoint: Entrypoint, db: AsyncSession
+) -> list[Garden] | None:
+    garden_entrys = await db.execute(
+        select(Garden)
+        .join(gardens_entrypoints, Garden.id == gardens_entrypoints.c.garden_id)
+        .where(gardens_entrypoints.c.entrypoint_id == entrypoint.id)
+    )
+    return garden_entrys.all()
```
OwenPriceSkelly, line 64: I don't think this is actually a `list[Garden]` -- [docs](https://docs.sqlalchemy.org/en/20/core/connections.html#sqlalchemy.engine.Result.all) say this is a sequence of `Row` objects, not `Garden` models. I think using `db.scalars(...` instead of `.execute` here would get us what you intended

OwenPriceSkelly: though it's kinda fun that duck typing let us get away without getting bit here

garden-backend-service/tests/api/routes/test_gardens.py
```diff
@@ -22,7 +23,9 @@ async def test_add_garden(
     create_entrypoint_with_related_metadata_json,
     create_shared_entrypoint_json,
     create_garden_two_entrypoints_json,
+    mocker,
 ):
+    mocker.patch("src.api.routes.gardens.BackgroundTasks", return_value=MagicMock())
```
OwenPriceSkelly: why is this the only place we need to mock the background tasks?

garden-backend-service/src/api/routes/entrypoints.py
```diff
@@ -50,16 +52,30 @@ async def get_entrypoint_by_doi(
...
         await db.delete(entrypoint)
         try:
             await db.commit()
+            if gardens:
+                for garden in gardens:
+                    db.refresh(garden)
+                    if settings.SYNC_SEARCH_INDEX:
+                        # just update the gardens, we don't want to delete a garden just because we deleted an entrypoint
+                        logger.info(
+                            msg=f"Deleting garden {garden.doi} from search index"
```
OwenPriceSkelly:
```suggestion
                            msg=f"Updating garden {garden.doi} on search index"
```

OwenPriceSkelly (approved):
LGTM! This has been a long time coming, really excited to see this get merged 🙌

I had a couple of nits/tweaks that we'll want to sanity check before we deploy it but otherwise I think this is in a great spot

Issue #111, opened 2024-07-03
feat: GET routes to query for multiple entities


OwenPriceSkelly:
We want the API to support things like "get a list of gardens belonging to a given user" etc

## Acceptance 
- GET /gardens and GET /entrypoints routes implemented 
- routes should accept query parameters to filter results

PR #118, opened 2024-07-15
get multiple entrypoints 

OwenPriceSkelly:
half of #111 

## Overview

now you can GET /entrypoints with query parameters to filter a result set, instead of only being able to fetch one at a time by doi. 

## Discussion

Slightly opinionated interpretation of list query parameters as OR filters rather than AND. This was because (since no entrypoint has more than one DOI) specifying multiple dois like `dois=doi1&dois=doi2&dois=doi3` clearly should mean "entrypoints with doi1 OR doi2 OR doi3". Then I wanted the query parameters to be consistent, so if you specify `tags=tutorial&tags=chemistry`, this grabs entrypoints with _either_ of those tags, not entrypoints with both. 

However, combined query parameters behave like AND -- so `dois=doi1&dois=doi2&dois=doi3&tags=tutorial&tags=chemistry` returns entrypoints which have "any of these 3 dois AND either of these two tags" 

## Testing

unit tests and some manual sanity checks

### DIFF
--- a/garden-backend-service/src/api/routes/entrypoints.py
+++ b/garden-backend-service/src/api/routes/entrypoints.py
@@ -73,7 +73,7 @@ async def get_entrypoints(
     if authors:
         stmt = stmt.where(Entrypoint.authors.overlap(array(authors)))
     if owner:
-        stmt = stmt.where(Entrypoint.owner.identity_id == owner)
+        stmt = stmt.join(Entrypoint.owner).where(User.identity_id == owner)
     if draft is not None:
         stmt = stmt.where(Entrypoint.doi_is_draft == draft)
     if year:
--- a/garden-backend-service/src/api/routes/entrypoints.py
+++ b/garden-backend-service/src/api/routes/entrypoints.py
@@ -56,24 +56,24 @@ async def get_entrypoint_by_doi(
 async def get_entrypoints(
     db: AsyncSession = Depends(get_db_session),
     *,
-    dois: list[str] | None = Query(None),
+    doi: list[str] | None = Query(None),
     tags: list[str] | None = Query(None),
     authors: list[str] | None = Query(None),
-    owner: UUID | None = Query(None),
+    owner_uuid: UUID | None = Query(None),
     draft: bool | None = Query(None),
     year: str | None = Query(None),
     limit: int = Query(50),
 ) -> list[EntrypointMetadataResponse]:
     """Fetch multiple entrypoints according to query parameters."""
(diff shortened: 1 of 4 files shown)


garden-backend-service/src/api/routes/entrypoints.py
```diff
@@ -49,6 +52,37 @@ async def get_entrypoint_by_doi(
...
+    db: AsyncSession = Depends(get_db_session),
+    *,
+    doi: list[str] | None = Query(None),
+    tags: list[str] | None = Query(None),
+    authors: list[str] | None = Query(None),
+    owner_uuid: UUID | None = Query(None),
+    draft: bool | None = Query(None),
+    year: str | None = Query(None),
+    limit: int = Query(50),
+) -> list[EntrypointMetadataResponse]:
```
hholb: Do we want to set a max limit? Relevant [docs](https://fastapi.tiangolo.com/reference/parameters/#fastapi.Query)
```suggestion
    limit: int = Query(50, le=100),
) -> list[EntrypointMetadataResponse]:
```

garden-backend-service/tests/api/routes/test_entrypoints.py
```diff
@@ -137,3 +139,163 @@ async def test_update_entrypoint(
...
+    mock_entrypoint_create_request_json,
+    mock_auth_state,
+    mock_auth_state_other_user,
+):
+    # ensure both users in DB
+    app.dependency_overrides[authenticated] = lambda: mock_auth_state_other_user
+    _ = await client.get("/greet")
+    app.dependency_overrides[authenticated] = lambda: mock_auth_state
+    _ = await client.get("/greet")
+
```
hholb, line 234: This is what I need for my uuid test! I was struggling to think of a way to add garden owned by a different user. I'm stealing this!

hholb (approved):
LGTM! :rocket:

garden-backend-service/tests/api/routes/test_entrypoints.py
OwenPriceSkelly: by all means! 🎁

garden-backend-service/src/api/routes/entrypoints.py
OwenPriceSkelly: ah yeah that's a good call. 100 seems sane, easy to change later if needed

WillEngler (approved):
Seems reasonable to me 👍

PR #121, opened 2024-07-17
Create routes for user's saved gardens

### LINKED ISSUE #116: feat: route(s) to allow users to add a garden to their list of saved_gardens
As a part of #105, #113 we added a field to the users for tracking saved gardens. We do not currently have routes that allow updates to this field. Something like `/users/{id}/saved_gardens` with GET, POST, and PATCH endpoints.

hholb:
Resolves #116 

## Overview

This PR adds new routes to manage a user's list of saved gardens. The new route is at `/users/{user_uuid}/saved/gardens` and supports GET, PUT, and DELETE requests.

## Testing
Unit tests
Manual testing

### DIFF
--- a/garden-backend-service/src/api/routes/users.py
+++ b/garden-backend-service/src/api/routes/users.py
@@ -1,7 +1,8 @@
 from typing import Any
 
-from fastapi import APIRouter, Depends, status
-from sqlalchemy import select
+from fastapi import APIRouter, Depends, HTTPException, status
+from sqlalchemy import insert, select
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from src.api.dependencies.auth import authed_user
 from src.api.dependencies.database import get_db_session
@@ -58,3 +59,40 @@ async def update_user_info(
         saved_garden_dois=saved_garden_dois,
     )
     return user_response
+
+
+@router.get("/gardens/saved")
+async def get_saved_garden_dois(
+    authed_user: User = Depends(authed_user),
+    db: AsyncSession = Depends(get_db_session),
+) -> list[str]:
+    return await _get_saved_garden_dois(authed_user, db)
+
+
+@router.patch("/gardens/saved")
+async def save_garden(
+    doi: str,
+    authed_user: User = Depends(authed_user),
+    db: AsyncSession = Depends(get_db_session),
+) -> list[str]:
+    user: User = await User.get(db, identity_id=authed_user.identity_id)
+    garden: Garden | None = await Garden.get(db, doi=doi)
+
+    if garden is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"No garden with doi {doi} found.",
+        )
+
+    try:
+        stmt = insert(users_saved_gardens).values(user_id=user.id, garden_id=garden.id)
+        await db.execute(stmt)
+        await db.commit()
+        dois = await _get_saved_garden_dois(authed_user, db)
+        return dois
+    except IntegrityError as e:
+        await db.rollback()
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="User has already saved this garden.",
+        ) from e
--- a/garden-backend-service/src/api/routes/users.py
+++ b/garden-backend-service/src/api/routes/users.py
@@ -79,7 +79,9 @@ async def get_saved_gardens(
 
     # We need to explicitly load the saved_gardens field
     await db.refresh(user, ["saved_gardens"])
-    return [GardenMetadataResponse.from_orm(garden) for garden in user.saved_gardens]
+    return [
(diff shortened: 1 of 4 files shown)


OwenPriceSkelly (changes requested):
This is close to what we'll want for the most part, but I think we'll want to expose the interface differently (my bad for not providing more detail in a ticket for you). 

  - I think we shouldn't just implicitly reach for the authed_user only; I think the routes here could be structured like `/users/{user_uuid}/saved/gardens` instead of just `/users/gardens`
  - this makes it easy to generalize to e.g. `/saved/entrypoints` in the future
  - `GET /users/{user_uuid}/saved/gardens` could return the full list of gardens instead of just dois (I think we still want only the list of dois in routes returning `UserMetadataResponse` though)
  - `PUT /users/{user_uuid}/saved/gardens/{doi}` to save a garden
  - `DELETE /users/{user_uuid}/saved/gardens/{doi}` to remove one 
  - also, all of these would be idempotent (we should still raise a 404 if a garden with that doi doesn't exist though)

OwenPriceSkelly:
Also, just for the sake of consistency, if we can get away with updating relationships via ORM-style sqlalchemy instead of core sqlalchemy I think we should (i.e. by updating a `user.saved_gardens` instance attribute instead of building insert/delete statements for the association table directly)

garden-backend-service/src/api/routes/users.py
```diff
@@ -62,71 +63,124 @@ async def update_user_info(
...
+    if user is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"No user with uuid {user_uuid} found.",
+        )
 
-@router.patch("/gardens/saved/{doi:path}")
+    # We need to explicitly load the saved_gardens field
+    await db.refresh(user, ["saved_gardens"])
+    return [GardenMetadataResponse.from_orm(garden) for garden in user.saved_gardens]
```
OwenPriceSkelly: I also like pydantic v1's `from_orm` name better for this method, but for the sake of consistency/not getting a bunch of deprecated warnings let's use their preferred method name
```suggestion
    return [GardenMetadataResponse.model_validate(garden) for garden in user.saved_gardens]
```

garden-backend-service/src/api/routes/users.py
```diff
@@ -62,71 +63,124 @@ async def update_user_info(
...
+
     if garden is None:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
             detail=f"No garden with doi {doi} found.",
         )
 
-    stmt = insert(users_saved_gardens).values(user_id=user.id, garden_id=garden.id)
+    await db.refresh(user, ["saved_gardens"])
+    for saved_garden in user.saved_gardens:
```
OwenPriceSkelly: I don't mind either way (up to you whichever syntax you think is clearer) but in case you're interested, I believe that we could take advantage of our `AsyncAttrs` mixin here like 
```python
for saved_garden in await user.awaitable_attrs.saved_gardens: 
    ...
```

OwenPriceSkelly: that being said, I don't think we actually want this loop, because we want PUTs to be idempotent; saving a garden you've already saved should be a no-op. Something like this probably "just works" with the orm, though I haven't tested it 
```python
await db.refresh(user, ["saved_gardens"])
if garden not in user.saved_gardens:
    user.saved_gardens.append(garden) 
```

garden-backend-service/src/api/routes/users.py
```diff
@@ -62,71 +63,124 @@ async def update_user_info(
...
         )
 
-    stmt = delete(users_saved_gardens).where(
-        and_(
-            users_saved_gardens.c.user_id == user.id,
-            users_saved_gardens.c.garden_id == garden.id,
+    await db.refresh(user, ["saved_gardens"])
+    updated_saved_gardens = list(filter(lambda g: g != garden, user.saved_gardens))
+    if len(updated_saved_gardens) == len(user.saved_gardens):
+        raise HTTPException(
```
OwenPriceSkelly: similar to the PUT method, DELETEs should be idempotent too -- I think we probably just want something totally naive here like 
```python
await db.refresh(user, ["saved_gardens"])
if garden in user.saved_gardens:
    user.saved_gardens.remove(garden)
```
without throwing an error if it's not saved / already deleted

garden-backend-service/src/api/routes/users.py
OwenPriceSkelly: though since we specified a response_model, fastapi would also do this conversion automatically:

```suggestion
    return user.saved_gardens
```

OwenPriceSkelly (changes requested):
This is looking very clean, good stuff -- only gripes are that I think we we should have PUT/DELETE be idempotent instead of throwing errors if the garden has already been saved/deleted. Besides those quibbles though this looks great!

OwenPriceSkelly (approved):
LGTM!

PR #122, opened 2024-07-18
Add `Project = "Garden"` to resource tags

### LINKED ISSUE #52: Infra: Add tags to all Garden resources

hholb:
Resolves #52 

## Overview

This PR adds  a tag `Project = "Garden"` to all current terraform resources that take a tag argument.

## Testing

`terraform plan` shows resources will be updated in place to add the new tag. No resources need to be replaced.


OwenPriceSkelly (approved):
LGTM! just applied on dev without a hitch 🚀

PR #125, opened 2024-07-23
MDF datasets table

### LINKED ISSUE #139: MDF datasets table
Create a table `mdf-datasets` that tracks the metadata for MDF datasets in the garden backend. This table will store only the minimum metadata needed to keep track of a dataset and any accelerate metadata, such as connected entrypoints and owner identity id. We will use this table to augment MDF searches with accelerate metadata.

MaxTuecke:
## Overview

Adds in the table `mdf_datasets` to our RDS. Will be used to store references to MDF datasets so we can augment them later on with analytics or linked entrypoints.


garden-backend-service/src/models/mdf_dataset.py
```diff
@@ -0,0 +1,20 @@
+from sqlalchemy import ForeignKey, String
+from sqlalchemy.dialects.postgresql import ARRAY
+from sqlalchemy.orm import Mapped, mapped_column
+from src.models.base import Base
+
+
+class Dataset(Base):
+    __tablename__ = "mdf-datasets"
+    id: Mapped[str] = mapped_column(primary_key=True)  # mdf SI subject id
```
OwenPriceSkelly: I think we should stick with the convention of having the `id` primary key be just arbitrary integers like the other tables. Not something I've personally been burned by before, but I buy into the [general wisdom](https://blog.ploeh.dk/2024/06/03/youll-regret-using-natural-keys/) on that side of the debate. I think having an `mdf_subject_id` field with a uniqueness constraint is the sweet spot (analogous to how we use `doi` for all the garden/entrypoint queries or `identity_id` for users, but the primary key for relationships is still just an integer)

garden-backend-service/src/models/mdf_dataset.py
```diff
@@ -0,0 +1,20 @@
...
+    user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
+
+    flow_action_id: Mapped[str]
+
+    doi: Mapped[str]
+    source_id: Mapped[str]
+    version: Mapped[str]
+
+    previous_versions: Mapped[list[str]] = mapped_column(
```
OwenPriceSkelly: we should sync up on this and decide if we want to normalize this and make versions into its own table with a 1-many relationship or leave as an array -- a rule of thumb for using an array type vs a related table is whether or not we can always treat the array value in the field as a single entity vs if we may sometimes be interested in the subvalues of the array

garden-backend-service/src/models/mdf_dataset.py
```diff
@@ -0,0 +1,20 @@
+from sqlalchemy import ForeignKey, String
+from sqlalchemy.dialects.postgresql import ARRAY
+from sqlalchemy.orm import Mapped, mapped_column
+from src.models.base import Base
+
+
+class Dataset(Base):
+    __tablename__ = "mdf-datasets"
```
OwenPriceSkelly: pretty arbitrary but we've been snake_casing our table names so far
```suggestion
    __tablename__ = "mdf_datasets"
```

garden-backend-service/src/models/mdf_dataset.py
```diff
@@ -0,0 +1,20 @@
...
+    id: Mapped[str] = mapped_column(primary_key=True)  # mdf SI subject id
+    user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
```
OwenPriceSkelly: this is for ownership, right? take a look at the extra ORM stuff we do for an `owner` field on gardens and entrypoints, should make the actual code for the queries a little easier to write

garden-backend-service/src/api/schemas/mdf_dataset.py
```diff
@@ -0,0 +1,13 @@
+from pydantic import Field
+
+from .base import BaseSchema, UniqueList
+
+
+class DatasetMetadata(BaseSchema):
```
OwenPriceSkelly: this might get confusing since we already have a `_DatasetMetadata` schema (only used by entrypoints so far) , so let's disambiguate 
```suggestion
class MDFDatasetMetadata(BaseSchema):
```

OwenPriceSkelly (changes requested):
Got some nitpicks but I think this is the right idea!

garden-backend-service/src/models/mdf/dataset.py
```diff
@@ -0,0 +1,29 @@
...
+    source_name: Mapped[str]
+    version: Mapped[str]
+
+    user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
+    user: Mapped[User] = relationship(lazy="selectin")
+    owner: Mapped[User] = synonym("user")
+
+    flow_action_id: Mapped[str]
+
+    previous_versions: Mapped[list[str]] = mapped_column(ARRAY(String), default=[])
```
OwenPriceSkelly: might be cleaner to just make this nullable rather than an empty list by default
```suggestion
    previous_versions: Mapped[list[str] | None] = mapped_column(ARRAY(String))
```

garden-backend-service/src/models/mdf/dataset.py
```diff
@@ -0,0 +1,29 @@
...
+    doi: Mapped[str] = mapped_column(unique=True)
+    versioned_source_id: Mapped[str] = mapped_column(unique=True)
+    source_name: Mapped[str]
+    version: Mapped[str]
+
+    user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
+    user: Mapped[User] = relationship(lazy="selectin")
+    owner: Mapped[User] = synonym("user")
+
+    flow_action_id: Mapped[str]
```
OwenPriceSkelly: assuming this is a UUID we can use the UUID type instead of str

OwenPriceSkelly (approved):
LGTM! couple of nits occurred to me on field types but nothing blocking 🙌

PR #126, opened 2024-07-23
Handle search index synchronization failures

### LINKED ISSUE #123: Handle search index synchronization failures
In #115 we added logic to synchronize the search index with the postgres database.

The current implementation schedules a background task to synchronize the search index after a successful write to the database. However, we do not handle cases where writes to the search index fail. This results in inconsistencies between the database and search index.

hholb:
Resolves: #123 

## Overview

In #115 logic was added to keep the search index in sync with the database. There is no logic to handle the event where an update to the search index fails, which can lead to inconsistencies between the database and the search index. This PR handles these cases by logging failed updates and periodically retries them in the background.

## Discussion

This PR adds a new database table called `failed_search_index_updates` where failed updated are logged. A background task called `retry_failed_updates` is started using the application's [lifespan](https://fastapi.tiangolo.com/advanced/events/) parameter that periodically looks for records in the new table and retries the update. The failed update record is removed from the table if the update is successful. There is a limit on how many times a failed update will be retried, when an update hits this limit it is skipped, but remains in the table.

There are a few tests that are marked to skip:
- One is due to not being able to delete an entrypoint that has an associated garden in the 'gardens_entrypoints' table. Currently, deletes from the 'entrypoints' table do not cascade to the 'gardens_entrypoints' table resulting in foreign key constraint violations.
- Two other tests that should test that failed search index updates resolve correctly when the search index is available do not work due to how the background task is started. The 'retry_failed_updates' task is started using the FastAPI lifespan parameter which is outside of the dependency injection system. We can't override the settings object in the test suite for the `retry_failed_updates` task the same way we can for routes that depend on the settings object. This means that the `retry_failed_updates` task does not run during tests. There may be a solution for this, but I haven't found one yet.

## Testing

Lots of manual testing
Unit tests


garden-backend-service/src/api/dependencies/auth.py
```diff
@@ -62,3 +63,12 @@ async def authed_user(
...
         raise HTTPException(status_code=500, detail="Internal Server Error")
+
+
+def get_auth_client(
+    settings: Settings = Depends(get_settings),
+) -> globus_sdk.ConfidentialAppAuthClient:
+    """Create an AuthClient for the service."""
+    return globus_sdk.ConfidentialAppAuthClient(
+        settings.API_CLIENT_ID, settings.API_CLIENT_SECRET
+    )
```
MaxTuecke, line 74: Glad this is getting added, was going to need this for some of the MDF stuff.

garden-backend-service/src/models/failed_search_index.py
```diff
@@ -0,0 +1,18 @@
...
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.orm import Mapped, mapped_column
+from src.models.base import Base
+
+
+class FailedSearchIndexUpdate(Base):
+    __tablename__ = "failed_search_index_updates"
+
+    id: Mapped[int] = mapped_column(primary_key=True)
+    doi: Mapped[str | None]
```
MaxTuecke: I'm wondering if DOI should be unique? Perhaps I'm misunderstanding how this table is being used, but AFAICT, if an update fails multiple times for the same resource, its not making new entries in the table, only incrementing retry_count and updating the failure information. Also, I'm not seeing a case where the DOI could be None, but again, maybe I'm missing something.

garden-backend-service/tests/api/tasks/test_tasks.py
MaxTuecke, line 1: Thorough tests as always. I salute the test tsar!

garden-backend-service/src/api/tasks/tasks.py
MaxTuecke, line 1: Just to confirm I understand how everything here fits together let me try and summarize how I think this is working. Please let me know if I'm missing / incorrect about any design choices :)

A user submits a task that requires a SI update. This starts the background task for `schedule_search_index_update`
- If the task fails to update the SI, immediately the failure gets logged in the `failed_search_index_updates` table. The table keeps track of the number of times a resource has failed to sync w/ the SI and the failure information.
- If the task succeeds to update the SI, we want to check `failed_search_index_updates` for any previous failures of that record and delete them. 


When the backend starts up, the lifespan goes and starts the background task `retry_failed_updates`. 
- Every `RETRY_INTERVAL` period, grab all the failed tasks from `schedule_search_index_update` that don’t have a `retry_count` bigger than `MAX_RETRY_COUNT`
- For each, re-try the SI sync
-- If it succeeds, we remove the failure entry from the table 
-- If it fails, we update the failure information and increment `retry_count`. 

When the backend shuts down, cancel the `retry_failed_updates` background task.

MaxTuecke (approved):
LGTM!

garden-backend-service/src/models/failed_search_index.py
hholb: I don't have strong feelings either way, I figured since we are essentially pulling the DOI out of the garden table, it should already be unique, and the insert logic should just update an existing record if it finds a duplicate DOI in this table. I am all for adding a unique constraint if that makes more sense/is clearer.

As far as possibly being None, I don't think there is a case where that will happen. I tend to make everything nullable until I have a solid reason not to. My thinking is that it is better to accept a partial record with some missing data than to reject it wholesale because of a missing field. However, in this case it doesn't make sense to have a failed update record that doesn't have an associated DOI, so I think making the DOI not null makes sense here.

garden-backend-service/src/api/tasks/tasks.py
hholb: Yep, I think you've got it! Thank you for taking the time to look it over!

I am not sure that starting the background task with the lifespan parameter is the best place to do it, but I haven't found a cleaner way. My issue is that the lifespan is outside of the dependency injection system which makes it hard to test. Any ideas?

MaxTuecke: I think using lifespan is the correct choice. Looking through the FastAPI docs, it seems to be the best way to do startup / shutdown that require a shared task. As for testing, if I'm understanding correctly, we want to have a different lifespan function for unit tests(?). I did a bit of poking around and found this: https://github.com/fastapi/fastapi/discussions/10833#discussioncomment-7945189

Assuming I'm understanding the goal correctly, this might work:
```
from contextlib import asynccontextmanager

@asynccontextmanager
async def test_lifespan(app):
    print("Test startup event")
    yield
    print("Test shutdown event")

@pytest.fixture(scope="session") # I add the following fixture to configure asyncio in pytest
def anyio_backend():
    return "asyncio"

@pytest.fixture(scope="session")
async def client():
    async with test_lifespan(app):
        async with AsyncClient(transport=ASGITransport(app=app), base_url="http://test") as client:
            yield client
```
Im not sure entirely why, but you have to change the pytest marker of the actual tests from `@pytest.mark.asyncio` to `@pytest.mark.anyio` and its not working with the current `patch_globus_groups` fixture, but maybe this helps.

garden-backend-service/src/api/dependencies/database.py
```diff
@@ -13,3 +13,9 @@ async def get_db_session(settings=Depends(get_settings)) -> AsyncSession:
 
     async with AsyncSessionLocal() as db_session:
         yield db_session
+
+
+async def get_db_session_maker(settings=Depends(get_settings)) -> AsyncSession:
```
OwenPriceSkelly: couple questions about this one, mostly to clarify what needs to be different from the similar one above it
  - why make it an async function? if we're just calling it once on application startup I think it's ok to block on it before handling other requests  
  - why make it a dependency? Is there a reason we'd want to use it in a route function over the other one? If not, I think we should move it out of the `dependencies` module 
  - since we're not using a context manager, does the session get closed anywhere? what happens if it doesn't?

OwenPriceSkelly: coming back to this - I think the type annotation threw me off a little, since it's returning a class and not a class instance it should be `type[AsyncSession]` not `AsyncSession` 

(they also seem to support the extra fancy generics typing syntax if you wanted to do `async_sessionmaker[AsyncSession]`)

OwenPriceSkelly: I do still think this shouldn't be a dependency though if we won't use it in routes -- we could honestly just define it `main.py` along with the lifespan function since we only use it there

garden-backend-service/src/api/tasks/tasks.py
```diff
@@ -0,0 +1,247 @@
...
+async def schedule_search_index_update(
+    update_operation: SearchIndexOperation,
+    garden: Garden,
+    settings: Settings,
+    db: AsyncSession,
+    auth_client: ConfidentialAppAuthClient,
+) -> None:
+    """Schedule an update on the search index for the given garden."""
+    try:
+        match update_operation:
```
OwenPriceSkelly, line 41: thank you for reminding me we have match/case now, I've been stuck mentally in 3.8

OwenPriceSkelly: also great use of match/case

garden-backend-service/src/api/tasks/tasks.py
```diff
@@ -0,0 +1,247 @@
...
+        db, doi=update.doi
+    )
+    if existing_update is None:
+        # Create a new record
+        db.add(update)
+    else:
+        # Update the existing record
+        existing_update.retry_count += 1
+        existing_update.error_message = update.error_message
+        existing_update.operation_type = update.operation_type
```
OwenPriceSkelly, line 182: do we need to update last_attempt here too?

garden-backend-service/src/api/tasks/tasks.py
```diff
@@ -0,0 +1,247 @@
...
+        await _delete_from_search_index(update.doi, settings, db, auth_client)
+
+
+async def _log_failed_update(
+    update: FailedSearchIndexUpdate,
+    db: AsyncSession,
+) -> None:
+    """Log a failed update attempt."""
+    existing_update: FailedSearchIndexUpdate | None = await FailedSearchIndexUpdate.get(
+        db, doi=update.doi
```
OwenPriceSkelly, line 173: understanding check: we select based on just doi and not doi + operation_type because we don't want cases where there's a failed update and a failed delete for the same doi, right? if so I think that'd be a valid reason to add a uniqueness constraint to doi, if not I think we could select based on both doi and operation type here

garden-backend-service/src/api/tasks/tasks.py
```diff
@@ -0,0 +1,247 @@
...
+
+        await asyncio.sleep(settings.RETRY_INTERVAL)
+
+
+async def _process_failed_updates(
+    failed_updates: list[FailedSearchIndexUpdate],
+    db: AsyncSession,
+    settings: Settings,
+    auth_client: ConfidentialAppAuthClient,
+) -> (int, int):
```
OwenPriceSkelly:
```suggestion
) -> tuple[int, int]:
```
I think this is the typing syntax for tuples, not tuple literals

garden-backend-service/src/api/tasks/tasks.py
OwenPriceSkelly: just want to chime in and say that it's Very Cool to see the intern brain trust paying off like this! Really appreciate having this extra context / writeup while I'm reviewing too 🙌

garden-backend-service/src/config.py
```diff
@@ -51,6 +51,8 @@ class Settings(BaseSettings):
 
     GLOBUS_SEARCH_INDEX_ID: str
     SYNC_SEARCH_INDEX: bool
+    RETRY_INTERVAL: int
```
OwenPriceSkelly: style nit: can we call it `RETRY_INTERVAL_SEC` so it's extra clear?

OwenPriceSkelly: also good call including these as settings

garden-backend-service/tests/api/tasks/test_tasks.py
OwenPriceSkelly: ➕🫡

garden-backend-service/src/models/failed_search_index.py
OwenPriceSkelly: agreed with both of you, I think it's sane either way but here's my 2c:

If I'm understanding you right, you're saying "we're never going to insert more than one with a given doi anyways so it's moot", which I think is a totally reasonable claim to make. It also strikes me as an assumption we might rely on now, but which our future selves could forget if we need to modify/extend the inserting code. IMO we might as well tell the db it's unique just so we can safely forget about it

garden-backend-service/src/api/tasks/tasks.py
OwenPriceSkelly, line 1: I have rare style anti-nit: this whole module is really clean/"pythonic", which isn't always easy. we literally love to see it 🙌 

Specifically a fan of: the operation type enum (and the match/case that uses it), the thin exception subclass for update errors, the logging is a nice bonus, function names are descriptive and helpers are well-scoped, and really nice exception handling throughout. 

Keep it up!

OwenPriceSkelly (approved):
This is fantastic, really lovely work. 

I had some quibbles/confusion around the get_db_session_maker but nothing major; I think we can get easily get this merged today 🚀

OwenPriceSkelly:
> There are a few tests that are marked to skip:
> 
> * One is due to not being able to delete an entrypoint that has an associated garden in the 'gardens_entrypoints' table. Currently, deletes from the 'entrypoints' table do not cascade to the 'gardens_entrypoints' table resulting in foreign key constraint violations.

This is intentional! I wouldn't worry about that one 

> * Two other tests that should test that failed search index updates resolve correctly when the search index is available do not work due to how the background task is started. The 'retry_failed_updates' task is started using the FastAPI lifespan parameter which is outside of the dependency injection system. We can't override the settings object in the test suite for the `retry_failed_updates` task the same way we can for routes that depend on the settings object. This means that the `retry_failed_updates` task does not run during tests. There may be a solution for this, but I haven't found one yet.

I also think I wouldn't worry about this one -- frankly I think that any real bugs we might run into in this kind of system probably wouldn't be caught by unit tests anyway

PR #128, opened 2024-07-29
Enable archiving of registered garden/entrypoint DOIs

### LINKED ISSUE #124: feat: enable archiving un-deletable entities with registered DOIs
If a garden or entrypoint has a registered (non-draft) DOI, we currently do not allow it to be deleted by anyone, even the owner. 

For the frontend to implement [tombstone pages](https://github.com/Garden-AI/garden-frontend/issues/151), the owner of a registered findable garden or entrypoint needs to be allowed to archive it. Owners of draft garden/entrypoints should still be allowed to delete them as usual. 

For a first pass, this just means setting an "archived" or "hidden" field on the tables, as well as updating datacite via their api (see: [tombstone best practices](https://support.datacite.org/docs/tombstone-pages#how-do-i-create-a-tombstone-page)). 

other considerations: DELETE requests should still fail if the entity isn't a draft; the archiving behavior should be an explicit update to an entity which both already exists and has a registered DOI.

hholb:
Resolves: #124 

## Overview

This PR adds functionality to mark registered garden/entrypoint DOIs as archived to support tombstone pages on the frontend.

## Discussion

I added PATCH routes for both gardens and entrypoints that allow updating the entities by DOI. The PATCH routes also handle the logic of archiving entities if the update request sets the 'is_archived' field and the DOI is not in draft state.

## Testing

Integration tests


hholb:
Notes to myself before merging:

- Merge #126 before this one
- Update background task calls to use the functions in #126 
- Fix logic in patch routes -- check for invalid is_archived/doi_is_draft state before calling db.commit()

MaxTuecke (approved):
Everything here seems reasonable. LGTM!

PR #130, opened 2024-07-30
MDF Datasets and Entrypoints relationship

### LINKED ISSUE #139: MDF datasets table

MaxTuecke:
Creates a new many to many relationship between the Entrypoints and Datasets tables (i.e. `Entrypoints.connected_mdf_datasets` and `Datasets.connected_entrypoints`)
Adding a MDF dataset to an Entrypoint's `connected_mdf_datasets` will in turn add the Entrypoint to the Dataset's `connected_entrypoints` and vise versa. Similarly, removing from one will remove from the other.

Also creates a new associations table `entrypoints_mdf_datasets` that is used to keep track of the relationships. Removing a relationship from either `connected_mdf_datasets` or `connected_entrypoints` will also remove it from the associations table.

Deleting either an Entrypoints or a Dataset will also properly remove any relationships it had.

Tested manually for now, but as the datasets API gets built out, will add unit tests.


hholb (approved):
LGTM!

The new table looks good and the relationships make sense to me. No qualms on my end.

OwenPriceSkelly (approved):
LGTM! before you merge though, coordinate with hayden to make sure that the alembic migrations will work nicely together -- I think that whoever merges second should just recreate their respective alembic migration script

PR #133, opened 2024-08-01
hotfix: add alembic migration script

hholb:
Add alembic migration script that was mysteriously missing from #126


OwenPriceSkelly (approved):
LGTM! (again)

PR #135, opened 2024-08-01
fix: alembic migration give `is_archived` field default value of False

hholb:
Related #124

The alembic script is failing after merging #124 due not null violations for existing records.

Fixed by setting a default value in the migration script.


OwenPriceSkelly (approved):
LGTM!

PR #136, opened 2024-08-07
MDF Datasets Search

### LINKED ISSUE #138: SEARCH route for MDF datasets
We want to be able to augment an MDF Globus search with any accelerate metadata (connected entrypoints, owner identity id, etc..) stored in our `mdf-datasets` table. Rather than directly querying Globus search, the garden backend will have a new search route `/mdf/search` that MDF uses. The route accepts and runs a `GSearchRequest ` and then for each dataset found, adds any accelerate metadata we have to said records `gmeta`

MaxTuecke:
## Overview

Creates new POST route `/mdf/search`

## Details

The search route accepts a Globus `GSearchRequest` in the request body, runs it against the MDF search index and then for each dataset in the result, augments it with any accelerate metadata stored in the `mdf_datasets` table (connected_entrypoints, owner_identity_id, etc..). 
If the Globus search query returns an error, the route will just forward it. Otherwise it responds with a  `MDFSearchResponse`, which is a `GSearchResponse` that has the extra field `accelerate_metadata` added to its `gmeta`.

The route is set it up so that it's called in the exact same way as a regular Globus search query, so swapping it out with existing search calls in MDF should be simple. Advanced Globus search queries are POST, so I kept it as a POST as well.

Does not require user auth since MDF search index also does not require auth.
Does not support Globus search scroll queries.

## Testing

Tested manually and have some basic unit tests.


hholb (approved):
LGTM!

garden-backend-service/src/config.py
```diff
@@ -58,6 +58,10 @@ class Settings(BaseSettings):
     DB_PASSWORD: str
     DB_ENDPOINT: str
 
+    MDF_SEARCH_INDEX: str = (
```
OwenPriceSkelly: Even if it's not likely to change anytime soon, lets keep this unset in application code and put it in the env variables (`garden-backend-env-vars/{dev, prod}` in aws secrets). Also, is there a dev vs prod MDF search index we could use for testing?

garden-backend-service/src/api/routes/mdf/datasets.py
OwenPriceSkelly, line 1: went back and forth on this a bit, but I think I am going to push back on not having a schema for the routes here. For the request schema, we can take advantage of the globus search [autodocs](https://search.api.globus.org/autodoc/redoc) and download their openapi spec to use with the automatic [pydantic model generator](https://docs.pydantic.dev/latest/integrations/datamodel_code_generator/). This is the approach we used for the `/doi` routes which similarly mirror the datacite API. 

For the response, it would also be nice to have an `AccelerateDatasetMetadata` schema that indicates what fields we're adding, too. I'm imagining something still pretty basic here, e.g. 
```python
class AccelerateDatasetMetadata(BaseSchema):
    # whatever new fields we're collecting 

class MDFSearchResponse(BaseSchema, <the autogenerated search schema>):
    # inherits fields from the search schema 
    # augmented with our metadata 
    accelerate_metadata: AccelerateDatasetMetadata
```
But I think it would go a long way. 

I'm being a stickler about this because I think that aggressively validating data at the boundaries of the system is how we avoid the sort of cruft you've seen accumulate in the bowels of the actual MDF codebase, where they're _constantly_ checking for missing fields, which is error prone and (as I know you've experienced) hard to follow or reason about if you weren't there when it was written. 

I'm noticing that most of the logic in this route is just checking to make sure that the data has the expected shape, which imo is the wrong place for that logic. Even just feeding the `_query_search` response into the appropriate pydantic model I think would simplify the body of your for-loop significantly here; e.g. we wouldn't need to remember things like "`len(entries)` is always 1" if it were guaranteed by the schema

OwenPriceSkelly (changes requested):
This is a solid first pass, and if I weren't a stickler about schemas this would be an easy LGTM, but alas I am one. 

Other small nits -- 1. can we have the module name match the route prefix (`/mdf/search.py` for the `/mdf/search` route); and 2. I think this PR should have at least a perfunctory unit test with some fake MDF/search data for the new route, rather than waiting until they're all implemented to write the tests

garden-backend-service/tests/api/routes/mdf/test_search.py
OwenPriceSkelly, line 1: tests look good! really helpful to see the sample payloads mocked out like that too.

garden-backend-service/src/api/schemas/mdf/dataset.py
```diff
@@ -1,12 +1,28 @@
...
+
+class MDFModernResult(BaseSchema, ModernResult):
+    accelerate_metadata: AccelerateDatasetMetadata | None = None
+
+
+class MDFLegacyResult(BaseSchema, LegacyResult):
+    accelerate_metadata: AccelerateDatasetMetadata | None = None
+
+
+class MDFGMetaResult(BaseSchema, RootModel[MDFLegacyResult | MDFModernResult]):
```
OwenPriceSkelly, line 23: Love these! I think using a RootModel to capture the legacy and modern result schemas together is a nice touch, too

OwenPriceSkelly (approved):
LGTM! 

thanks for humoring me on the schemas, they turned out great 🌱

PR #143, opened 2024-09-05
Add new DB models for Modal functions and apps

WillEngler:
Starting on Garden <-> Modal integration.

## Overview

This PR adds new DB models and tables. The new models are ModalApp and ModalFunction. We need to model ModalApps
because they are Modal's unit of deployment, but we won't track them in DataCite. ModalFunctions are similar to Entrypoints - Gardens can collect them, they will get DOIs, and users can invoke them from the SDK. This PR also adds 

## Discussion

This is my first time in the new shiny Garden backend, so I may be doing some obvious things wrong with the new models. I was able to work off existing examples, but I wouldn't be surprised if I'm doing something silly.

## Testing

No tests in this PR. I think we need to get a PUT /modal-app endpoint going to have functionality to test.


OwenPriceSkelly (approved):
LGTM!

PR #144, opened 2024-09-12
Owen/better logging

### LINKED ISSUE #141: enhancement: better logging practices
It's more difficult than it needs to be to diagnose issues like #140 with our current logging setup. 

We have two buckets of work here: "Logging discipline" in the codebase itself, and logging infra outside of the codebase. This issue is for the first bucket of work. 

We don't really have much of an existing convention for logging. Even if we had state-of-the-art industrial strength logging infrastructure, our existing code wouldn't really take advantage of it.

## Acceptance criteria 
  - Our logging should play nice with the async nature of the app -- I should be able to see the logs for a given request without logs from unrelated requests interleaved
  - We should use DEBUG/INFO/WARNING/ERROR log levels appropriately 
    - Also, exceptions should be logged with full stacktraces 
  - We should include other possibly-useful context like timestamps, user ids, etc 
  - NB: We don't want the app to do anything but stream logs to stdout (https://12factor.net/logs)
  - Something like [structlog](https://www.structlog.org/en/stable/why.html) is maybe also worth adopting on top of the builtin python logger

OwenPriceSkelly:
closes #141  
## Overview

This should be a huge improvement over our current logging setup, but the configuration I've landed on here with structlog is not meant to be super opinionated, I think we should be actively looking for things to change / improve so our logs are actually useful. 

#### New modules: 
  - `src.logging`: currently just configures structlog and the builtin logger to play nicely with structlog. if we want to write something like a custom exception processor for logs in the future it would live here.
  - `src.middleware.logging`: defines two fastAPI (starlette) middleware classes to include a request correlation uuid and include a final "request processed" log with the time elapsed as well as method/path/status code etc.
  
  
## Discussion
For the most part, logging with structlog looks a lot like the standard library logging with a few differences. It's worth checking out the full [docs](https://www.structlog.org/en/stable/why.html), but here are some thoughts on tentative "best practices" / patterns to get us started: 

  - there should be a `logger = structlog.get_logger(__name__)` line at the top level of the python module you're working in (not inside the route function). 
  - anywhere else in the module (e.g. inside of a route function), you should create a new logger from the module-level one. This is nice to pre-fill logs with additional "local" context like a doi: 
  ```python
  logger = structlog.get_logger(__name__)
  ...
  def some_route(doi, authed_user, etc, ...): 
      route_logger = logger.bind(doi=doi) 
      ...
      route_logger.info("something happened")  # automatically includes doi in the log
      logger.warning("uh oh") # top level logger does not include doi 
  ```
  - for """global""" context we want to provide every logger (module-level or local) we need to tactically use the structlog contextvars integration, like `with structlog.contextvars.bound_contextvars(...)` (see the changes to authed_user for example). This is how we get values to appear in all logs emitted over the course of a single request / event in the async event loop, like a username/id or a request correlation id.
    - This actually played very nicely with fastapi out of the box as long as I didn't try using it outside of a dependency or middleware 
    - see also: https://www.structlog.org/en/stable/contextvars.html 
  - avoid using the builtin logging library, since those logs won't have any special configuration 

## not in this PR:
  - no special formatting for exception logs. a plain `log.exception()` in a regular `except` block is typically probably useful and will dump the stacktrace, but there won't be any of the special formatting structlog could do for us. I started toying around with it but didn't see an easy way to get the logs to stop including `Settings` objects (with all of our aws secrets) in the plain text output 
  
## Testing

just manual tests, but it seems like structlog has pretty good support for testing that logs have expected content etc if we decide we need it.


WillEngler (approved):
Seems like a big improvement! Thanks for systematizing the way we do this and identifying a good library to build on

PR #155, opened 2024-09-19
Full-text search with postgres

### LINKED ISSUE #154: Prototype full-text search in Postgres
How much of a lift would it be to stand up a good-enough replacement for the search index just using Postgres? This ticket is for a "research spike" - spend no more than a week on this and see where we're at. The goal is to make a new version of GET /gardens that takes a text snippet from a search bar and filter fields and returns a list ordered by relevance. Full text search should be done over the contents of the garden and its child entrypoints. We do not need to implement faceting.

Broadly, there are two outcomes here:
- We learn this is pretty straightforward (only takes a few days to make a route that's good enough). We cut additional tickets to productionize it, change the frontend to use it, and get rid of the search index.
- We learn that it's decently involved and the whole project would take a month or so. We produce an artifact with what we've learned that we can come back to later if the need to move away from the search index proves to be urgent/pressing.

hholb:
Resolves: #154 #161 

## Overview

This PR adds a new route `POST /gardens/search` supporting full text search and filtering of garden records. The route also returns facet counts based on applied filters, and supports paginating results using the offset parameter in the request body.

## How the route works
In the current state `/gardens/search` takes a request body with a search query and optional list of filters, for example:
```
{
  "q": "material science",
  "limit": 10,
  "offset": 5,
  "filters": [
    {
      "field_name": "authors",
      "values": [
        "will engler"
      ]
    }
  ]
}
```
The filters are applied as added `WHERE` clauses to the SQL query which means multiple filters are ANDed together.

The response is a `GardenSearchResponse` that looks like:
```
{
  "count": 5,
  "total": 5,
  "offset": 0,
  "garden_meta": [<garden metadata>, ...],
  "facets": {
    "tags": {
      "some tag": 5
    },
    "authors": {
      "some author": 5
    },
    "year": {
      "2023": 2,
      "2024": 3
    }
  }
}
```

## How the text search works
The text search and ranking happens in an custom SQL function. I tried to map the function to sqlalchemy statements but ran into errors trying to call the postgres `setwieight` function through sqlalchemy. I decided on registering the function with the postgres server from raw sql using a new `init` and its async cousing `async_init` in the database dependencies.

The actual text search is pretty straight forward after spending some time to grok postgres [full text search docs](https://www.postgresql.org/docs/current/textsearch.html). Essentially we need to concatenate the text fields we are searching into a single 'document' using `to_tsvector` and the `||` operator, and assign relative weights (A, B, C, D) to each part of the document, e.g. authors:A, contributors:B, description:C etc. Fields weighted with A have higher relevance than B, and so on.

Assigning weights is accomplished with the built in `setweight` function in postgres like:
```
SELECT setweight(to_tsvector(description), 'A') ||
       setweight(to_tsvector(title), 'B') as weighted_document
FROM gardens;
```

The `to_tsvector` function tokenizes the the text and does some basic lexical manipulation to prepare for searching.

 Once we have the weighted documents, it is easy to get a list of matching gardens ranked by relevance:
```
SELECT id
             ts_rank(weighted_docuemnt, query) AS rank
FROM gardens
WHERE garden_document @@ query; -- @@ returns true is the query is in the document
```

## Testing
Manual testing with a local database populated with gardens and entrypoints from the dev search index.

New pytest tests that exercise this route.


garden-backend-service/tests/conftest.py
```diff
@@ -71,30 +74,48 @@ def db_url(pg_container) -> str:
...
+    yield
+
+    # Clean up after the test
+    with Session(_sync_engine) as db:
+        db.execute(text("DROP TABLE gardens_entrypoints;"))
+        db.execute(text("DROP TABLE entrypoints CASCADE;"))
+        db.execute(text("DROP TABLE gardens CASCADE;"))
+        db.execute(text("DROP TABLE users CASCADE;"))
+        db.execute(text("DROP TABLE failed_search_index_updates CASCADE;"))
+        db.commit()
```
WillEngler, line 117: Why move away from `Base.metadata.drop_all(engine)`? Is it because sqlalchemy doesn't know about the search SQL?

garden-backend-service/src/api/search/utils.py
```diff
@@ -0,0 +1,113 @@
...
+        func.count().label("count"),
+    ).group_by(column("author"))
+
+    year_query = select(
+        filtered_gardens.c.year.label("year"), func.count().label("count")
+    ).group_by(column("year"))
+
+    tags_result = await db.execute(tags_query)
+    authors_result = await db.execute(authors_query)
+    year_result = await db.execute(year_query)
```
WillEngler, line 108: If the performance impact is not material feel free to ignore, but is it possible to combine these into one query or run the queries in parallel to avoid three serial round trips to the DB?

garden-backend-service/src/api/search/sql.sql
```diff
@@ -0,0 +1,86 @@
...
+       FROM garden_documents gd
+       LEFT JOIN ep_ranks
+       ON ep_ranks.garden_id = gd.garden_id
+    )
+    SELECT gr.garden_id, gr.rank
+    FROM garden_ranks gr
+    WHERE gr.rank > 0
+    ORDER BY gr.rank DESC;
+END;
+$$ LANGUAGE plpgsql;
```
WillEngler, line 26: I always find CTEs a little mystifying, but this is nice and readable for all the work it's doing. Great stuff

garden-backend-service/src/api/routes/gardens.py
```diff
@@ -96,6 +99,54 @@ async def search_gardens(
...
+    facets = await calculate_facets(db, stmt)
+
+    # Get totals for offset/pagination
+    total_count_stmt = select(func.count()).select_from(stmt.subquery())
+    total_count = await db.scalar(total_count_stmt)
+
+    # Run the search query
+    result = await db.scalars(
+        stmt.limit(search_request.limit).offset(search_request.offset)
+    )
```
WillEngler, line 139: Similarly here, could we parallelize these requests? (like with asyncio.gather)

garden-backend-service/tests/api/routes/test_gardens.py
```diff
@@ -627,3 +627,191 @@ async def test_disallow_editing_published_garden_fields(
...
+    assert result["count"] == 10
+    assert result["offset"] == 0
+
+    body = {"q": "garden", "offset": 15}
+    res = await client.post("/gardens/search", json=body)
+    assert res.status_code == 200
+    result = res.json()
+    assert result["total"] == 20
+    assert result["count"] == 5
+    assert result["offset"] == 15
```
WillEngler, line 817: Nice tests!

WillEngler (approved):
Some things around parallelizing independent DB queries to consider, but overall this looks great

garden-backend-service/tests/conftest.py
hholb: Yeah, calling `drop_all` fails because sqlalchemy doesn't know about the views and `drop_all` doesn't have a `CASCADE` option.

garden-backend-service/src/api/search/utils.py
hholb: I had the same thought and tried this running these in an `asyncio.gather`. It needed some extra setup to work because the AsyncSession object is [mutable and not thread safe](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#session-faq-threadsafe). Basically, we need a unique session for each parallel query which is not too hard. I was able to get it working, but issuing the queries in parallel doubled average response time in my benchmark vs issuing them serially.

garden-backend-service/tests/api/routes/test_gardens.py
hholb: Thanks!

garden-backend-service/src/api/search/sql.sql
hholb: I love a good CTE! I'm sure there are some doozies out there, typically I think they help break complex queries down into smaller chunks that are easier to reason about.

garden-backend-service/src/api/search/utils.py
WillEngler: Interesting! Thanks for digging in to that

PR #156, opened 2024-09-25
first pass at POST /modal-invocations  

### LINKED ISSUE #150: New route that invokes Modal function in Garden workspace
Create a POST /modal-function-invocation route. This should take the ID of a published modal function and serialized arguments. It should execute the function from the Garden backend using Garden's Modal credentials and pass the serialized output back to the user.

This will integrate closely with the SDK and likely have a sibling PR in the SDK that builds on https://github.com/Garden-AI/garden/issues/534. 

This issue description is intentionally a bit underspecified. How should we handle serialization? What size limits on payloads should we impose? There are a lot of questions to work through here. I recommend getting a pretty naive first stab working and going from there.

OwenPriceSkelly:
closes #150 

## Overview

This PR is the first pass at a route which allows us to invoke modal functions via the backend on behalf of users. 

### What this doesn't do:
  - anything that requires talking to the DB (except indirectly by depending on `authed_user`). This just performs a naive function lookup with modal by function and app name as provided by the request body. The next pass at this would take the id of the function and fetch the fields from our DB before getting the function from modal. 
  - any usage tracking, either per-user or at the organization level. Next pass at this can make sure that a) we still have credits in the bank and b) that the user hasn't gone over their individual limit
  - any serializing/deserializing the raw input args or response from modal. The garden sdk will need to handle this in https://github.com/Garden-AI/garden/issues/534 before we've replicated the entire process end-to-end for invoking a function on modal directly.
  
### What this does do:
  - hides the route logic behind a MODAL_ENABLED feature flag (`MODAL_ENABLED` as well as `MODAL_TOKEN_{ID, SECRET}` environment variables will need to be set in aws) 
  - includes a new pydantic custom type `B64Bytes` which can be used as a drop-in replacement for plain `bytes` in a pydantic schema. That is, when accessing a `B64Bytes` field of a request body the value will be `bytes`, and you can leave a value in a response body as plain `bytes`, but the json representation that gets sent over the wire should be a b64 encoded `str`. 
    - tl;dr postman will see a base64 encoded string; the route function handling the payload will see plain bytes.
  - a `get_modal_client` dependency other modal routes can use 
    - NB: this is an async client, so we may need both a sync and async version of this dependency 
  - tries its very best to follow the exact same steps as the modal client library when it builds RPC requests for the modal backend. this meant accessing a few underscored functions/classes and duplicating a code path which I am 99% sure is not reachable, just in case I'm wrong. 
    - I think we'll be safe just pinning the exact same version of the modal library in the sdk as the backend, but there's a slim chance that they refactor/change something big and we'll need to update our logic here to stay in line with theirs. (for the record my impression from poking around their codebase/git history is that this is very unlikely)
   

## Testing
Mostly manually with the same script I demoed last week. 

I did add a basic happy-path unit test that mocks out anything which would actually talk to the modal server (and so we've got a fixture now for the modal client) but didn't bother with simulating the case where the remote function throws an exception and we need to return the full traceback (but I did test this manually and it ✨just works ✨)


WillEngler (approved):
LGTM!

PR #158, opened 2024-10-01
Restrict edits on published gardens/entrypoints

### LINKED ISSUE #149: feat: allow minor edits to published metadata
# Description
Currently, the backend naively treats "should this entity be editable" as equivalent to "should this be deletable" -- i.e. users can only edit gardens or entrypoints they could have deleted (see: places where the `assert_deletable_by_user` helper function is used). 

Turns out this approach was too restrictive; "minor edits" are generally fine, even for published entities we wouldn't want to allow users to delete.

## Recap: when can gardens or entrypoints be _deleted_?
a `DELETE` request should fail if _any_ of these conditions are met: 
  - the user making the request is not the entity's owner 
  - the doi has been archived with datacite (`is_archived=true`) 
  - the doi has been published with datacite (so `doi_is_draft=false`)

(This is the current behavior of the backend, should be no need to change the deletion logic)

## When can they be _modified_?
A `PATCH` request should fail if any of these conditions are met: 
  - the user making the request is not the entity's owner 
  - the doi has been archived (IMO archiving should mean "no longer changing", but open to pushback) 
  - the doi has been published, **AND** the edits described by the request body would be "major changes" 

### What constitutes "major changes"? 
Unlike a deletion, an update should only fail if the edits constitute "major changes" to the entity we've registered with datacite. 

For a garden, we've decided "major changes" would be the changing set of entrypoints it refers to. Any other metadata edits (title, authors, description etc) are minor enough to allow. 

For an entrypoint, "major changes" would be anything that might impact its behavior -- e.g. "lower level" metadata fields like `func_uuid, container_uuid, full_image_uri, short_name` etc. Any other edits (including changes to related materials like datasets) should be allowed.

## Other considerations
  - We currently allow edits through either a PATCH or a PUT -- for this issue, just the PATCH routes should be in scope 
  - After a successful edit to a non-draft entity, we should also update datacite through their API. The `archive_on_datacite` helper is pretty basic but probably useful reference for writing a similar `update_on_datacite` helper 
  - A request to archive or publish an entity hits the same PATCH routes, so we'll want to make sure that the archiving/publishing logic is handled separately to editing general metadata fields  
  
# Acceptance criteria:
  - users can make "minor edits" to the metadata of their published gardens or entrypoints

shane-jung:
## Overview

Adds assertion to ensure that edits are not made on archived Gardens/Entrypoints. 

Adds restrictions on which attributes can be edited on published entrypoints

Fixes issue #149

## Discussion

Currently a user is not able to edit the list of entrypoints associated with a garden. A follow up PR will add in that ability, still ensuring that only unpublished gardens can have their entrypoints changed.


OwenPriceSkelly (changes requested):
So far this is looking great! No nits with anything you've got so far, but let's add a couple of unit tests before we merge (thinking we should at least test that a valid edit is successful & that an illegal edit returns the correct error code). lots of good examples in the repo to get you started but also happy to sync up if you run into trouble, tests can be finicky

OwenPriceSkelly (approved):
LGTM! You can do the honors if you want to squash/merge this yourself 🫡 

P.S. welcome to the backend gang 🤝

PR #165, opened 2024-10-03
Persist Modal metadata

### LINKED ISSUE #148: Backend lets you fetch modal functions
Implement GET /modal-apps and GET /modal-functions.

For GET /gardens, garden objects will include child modal function

WillEngler:
Closes #148

## Overview

This PR makes it so that POST /modal-apps actually creates ModalApp and ModalFunction records in the DB. It also adds GET /modal-apps

## Discussion

I'm new to the system of DB models + Pydantic schemas and very open to feedback on how I'm laying them out.

## Testing

Tested manually with Postman. Also implemented GET /modal-apps in this PR for the purpose of having a unit test that retrieves the Modal app's metadata after POSTing it.


OwenPriceSkelly (approved):
LGTM! 

we synced up to review this one but just for bookkeeping: 
- cascade settings for sqlalchemy relationships are mysterious and we should do a deeper dive for our own understanding at some point 
- we'll need to explicitly handle the business logic for overwriting not-yet-published modal functions even before we worry about dois 
- we're punting on dois for modal functions for as long as it makes sense to do so

PR #168, opened 2024-10-04
Reject changes to entrypoint_ids on published Gardens

### LINKED ISSUE #167: feature: Allow updates to entrypoint_ids on unpublished Gardens
Currently, a garden's associated `entrypoint_ids` cannot be modified, regardless of what state the DOI is in. Users should have the ability to edit this list as long as a garden isn't published.

shane-jung:
Builds on #149 and #158, which restricted which fields could be edited on entrypoints

## Overview

Adds ability for unpublished gardens to have their entrypoints edited. Rejects changes to entrypoint_ids on published gardens. Adds an accompanying request to ensure changing this field on a published garden raises an error. 

## Discussion

This PR builds on functionality implemented in #149 that ensured that Gardens and Entrypoints could not be edited while in an archived state. 

## Testing

To test this, I submitted PATCH requests to unpublished and published gardens, with different combinations of fields changed. In particular, we should expect that any PATCH requests that include a change to entrypoint_ids fails *if* that garden is in a published state.  Conversely, we should expect that PATCH requests that include a change to entrypoint_ids succeeds when that Garden is in draft state. I also added an additional unit test that verifies this expected behavior.

### DIFF
--- a/garden-backend-service/src/api/routes/gardens.py
+++ b/garden-backend-service/src/api/routes/gardens.py
@@ -258,18 +258,15 @@ async def update_garden(
 
     garden_patch_dict = garden_data.model_dump(exclude_none=True)
 
-    # NOTE: This is commented out because currently entrypoint_ids are not part of the GardenPatchRequest,
-    # so any attempt to update entrypoint_ids will not change anything. If we want to allow updating entrypoint_ids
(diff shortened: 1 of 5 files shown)


WillEngler (approved):
Clean, readable, and the new test case is nice. LGTM!

PR #172, opened 2024-10-08
Run linter and formatter in CI and pre-commit

### LINKED ISSUE #170: chore: set up CI to use black, flake8 etc like sdk
haven't yet gotten around to setting up auto formatting for the backend like on the sdk side, but it's easy to forget to `poetry run black .` every time manually 

this could also be a good excuse to check out [ruff](https://github.com/astral-sh/ruff) as a (purportedly much faster) drop in replacement

hholb:
Closes #170 

## Overview

This PR adds `ruff` as a dev dependency and sets up our CI and `pre-commit` configs to run `ruff check` and `ruff format`. This PR removes `black` and `flake8` as dev dependencies because `ruff` is a drop in replacement for both.

All we need now is a faster `mypy`...

## Discussion

I ran `ruff check --fix` and it touched quite a few files, mostly cleaning up imports.
Much faster than black and flake8!

## Testing

Manual testing
Tested the CI job using `act`


.github/workflows/ci.yaml
```diff
@@ -52,15 +53,24 @@ jobs:
...
           cd ./garden-backend-service
-          poetry run pytest
+          poetry run safety check
 
-      - name: Check for vulnerable libraries with safety
+      - name: Format with Ruff
         run: |
           cd ./garden-backend-service
-          poetry run safety check
+          poetry run ruff format
```
OwenPriceSkelly: I think we actually don't want a format step in the action, just the lint step -- what happens if e.g. I don't have pre-commit hooks installed and submit an unformatted pull request?  
  1. code is unformatted on my branch 
  2. `Format with Ruff` step runs, the code is formatted on the worktree inside this GH action (but not in any commit on my branch)
  3. `Lint with Ruff` step runs, sees only the formatted code from step 2 and doesn't complain 
  4. Build succeeds and we merge from my unformatted branch
  
Also, philosophically, I don't think we would actually want a CI pipeline to change the code at all (even superficially) once it's been reviewed.

OwenPriceSkelly: we definitely do want the precommit hook to run the formatter though

OwenPriceSkelly (changes requested):
Thanks for picking this up! one small gotcha, but once we're just linting (and not formatting) in the GH action this is good to go!

.github/workflows/ci.yaml
hholb: That makes sense!

OwenPriceSkelly (approved):
LGTM! ([with flying colors](https://github.com/Garden-AI/garden-backend/actions/runs/11256098863))

PR #173, opened 2024-10-09
Delete modal apps and edit modal functions

### LINKED ISSUE #147: Update and delete Modal Functions
Create new routes, `DELETE /modal-functions` and `PATCH /modal-functions`. They should behave analogously to the /entrypoints routes with those verbs, and have the same rules around DOI state management.

WillEngler:
Closes #147

## Overview

This PR adds DELETE /modal-apps and PATCH /modal-functions routes. I also consolidate fields shared between the ModalFunction and Entrypoint schemas into parent CommonFunction classes.

## Discussion

There are a few subtle things going on here that I breezed through. One is different logic for how Modal Functions represent draft vs published state. We're thinking that to cut out some DOI munging, Modal Functions will only get a DOI once they are published. So their draft state is just having a DOI of none. If we like this, we'll probably want to extend it to entrypoints and gardens in time, but that will be a somewhat involved migration as lots of code assumes these entities always have DOIs.

Another noteworthy discrepancy between Modal Functions and Entrypoints is "short_name" vs "function_name". I've gone with "function_name" for modal functions because it more clearly describes the concept. We might want to standardize this. (Either both short_name or both function_name)

As these routes aren't on the critical path for getting this flow "Ben-ready" my focus here was just banging out the minimum functionality to have our main CRUD endpoints in place. I haven't explicitly tested things like archive logic. But i have made what I consider a reasonable stab at matching the analogous garden and entrypoint routes.

## Testing

I added unit tests for the main happy path of each route.


OwenPriceSkelly (approved):
LGTM! and definitely think that `function_name` is better than `short_name` -- honestly not sure where `short_name` came from but it might be a holdover from a dlhub schema

PR #174, opened 2024-10-09
add support for sorting search results by field

hholb:
## Overview

This PR adds support for sorting search results by field in the `/gardens/search` route.

The `GardenSearchRequest` now accepts an optional `sort` parameter that looks like:
```
{
 <other GardenSearchRequest fields...>,
  "sort": {
    "field_name": "title",
    "order": "desc"
  }
}
```
## Testing

Added new unit tests

### DIFF
--- a/garden-backend-service/tests/api/routes/test_gardens.py
+++ b/garden-backend-service/tests/api/routes/test_gardens.py
@@ -829,7 +829,9 @@ async def test_search_gardens_sort_by_title(
     sorted_titles = ["A", "B", "C", "D", "E", "F"]
 
     random_titles = list(sorted_titles)
-    random.shuffle(random_titles)
+    # shuffle until we know the lists are different
+    while random_titles == sorted_titles:
+        random.shuffle(random_titles)
 
     for title in random_titles:
         garden_data = deepcopy(mock_garden_create_request_no_entrypoints_json)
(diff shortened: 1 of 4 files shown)


WillEngler (approved):
I have a non-blocking nit on one of the tests, but overall this looks great

garden-backend-service/tests/api/routes/test_gardens.py
```diff
@@ -815,3 +816,73 @@ async def test_search_gardens_offset_paginates_results(
...
+async def test_search_gardens_sort_by_title(
+    client,
+    mock_db_session,
+    override_authenticated_dependency,
+    mock_garden_create_request_no_entrypoints_json,
+):
+    sorted_titles = ["A", "B", "C", "D", "E", "F"]
+
+    random_titles = list(sorted_titles)
+    random.shuffle(random_titles)
```
WillEngler: I'm not too worried about false positives here (there's only a 1 out of 6 factorial chance the list stays in place, and the asc vs desc logic mitigates that being a problem) but generally I don't think we should inject non-determinism into this kind of test

hholb: Good catch! I don't want to think too hard about this, but a bit of a crude solution is to do:
```
while random_titles == sorted_titles:
    random.shuffle(random_titles)
```
I suppose there is a small chance of an infinite loop, but I think the average number of iterations will be <2

WillEngler: Ha, I like it

Issue #159, opened 2024-10-02
allowlist garden devs for `/modal-*` routes 


OwenPriceSkelly:
We want the routes to upload and invoke modal apps/functions to be live and usable by us but not the general public. 

This could stub out helpers that #151 can later implement, but the logic should be "is this user ben or a garden dev" and not "has this user reached their quota"

WillEngler:
Closed by #175

PR #175, opened 2024-10-09
Restrict Modal routes to allow-listed users

WillEngler:
Closes issue #159 

## Overview

The app will fetch a list of VIPs (for now, the dev team) and make sure that only users with email addresses in that list can access Modal routes.

## Testing

I tested this locally by playing with the MODAL_VIP_LIST in my .env. I tried it with my email in it and could access Modal routes. I removed it and I couldn't.


OwenPriceSkelly (approved):
LGTM 🎩🕴️

PR #176, opened 2024-10-10
include middleware to log unhandled exceptions

OwenPriceSkelly:
## Overview

Quick PR to get us better logs for unhandled exceptions besides HTTPExceptions 

## Discussion

Turns out we don't get logs at all besides a 500 status code if the exception our app throws isn't an HTTPException -- this middleware just catches (typically unexpected) other exception types and includes a stacktrace we can peruse in the lightsail console 

## Testing

just manually on my local dev server so far


WillEngler (approved):
The more logs the better 🪵

PR #181, opened 2024-10-14
Tweaks to Modal metadata routes

WillEngler:
Doing this while I'm fixing up Shane's frontend branch. Two things:

- The `id` field in Modal Apps and Modal Functions wasn't showing up in our auto-gen'd OpenAPI schema. `id` is treated specially apparently. We want that field in there for the frontend to have in its types.
- I removed the Modal VIP gate from GET /modal-apps and GET /modal-functions. This lets us fetch and show Modal Functions in the garden view without worrying about 500s for non-VIPs.


OwenPriceSkelly (approved):
LGTM 🚢

Issue #151, opened 2024-09-16
Per-user usage tracking for Modal invocations


WillEngler:
A core aspect of us letting users "preview" Modal functions using our credits is that we need to meter and limit usage. By counting execution time paired with type and number of processors, we should be able to approximate usage cost per user and cut off invocations after a user passes a certain threshold. (Say, $5/month)

PR #182, opened 2024-10-15
First pass at Modal usage tracking

hholb:
#151

## Overview
 
This PR adds logic to the `invoke_modal_fn` route to do a rough calculation of billable usage specific to a user per invocation.

There is a new table called `modal_invocations` where this data is tracked. When a user makes a request to invoke a modal function a check is performed as an auth dependency, `under_modal_usage_limit` to make sure the user is under a monthly usage limit. The usage limit is a new settings option called `MODAL_USAGE_LIMIT`.

## Discussion
I am thinking since we are not planning on billing our users based on Modal usage, we are safe making a rough estimate of usage as long as we over-estimate on our end compared to the actual billable usage on Modal. 

### What this does
- Calculates a rough usage estimate per invocation based on the function's hardware spec and the time it takes from when we issue the request to Modal to when we get a response back. The calculation is based on Modal's [pricing](https://modal.com/pricing) per hardware resource multiplied by the number of seconds it takes to get a response.
- Rejects invocation requests when a user is over the limit

### What this doesn't do
- Account for queue time after making an invocation. This may be an issue if users are requesting a lot of GPUs.
- Enforce any timeouts or other checks on invocations

## Testing

A new unit test to make sure we reject invocations when a user is over the limit.


garden-backend-service/tests/api/routes/modal/test_invocations.py
```diff
@@ -81,3 +82,72 @@ async def test_invoke_modal_fn(
...
+    # Mock retry_transient_errors to avoid outbound network calls
+    mock_retry = mocker.patch("src.api.routes.modal.invocations.retry_transient_errors")
+    mock_retry.side_effect = [
+        MagicMock(
+            function_call_id="mock_call_id", pipelined_inputs=["mock_input"]
+        ),  # FunctionMap response
+        MagicMock(
+            function_call_id="mock_call_id", pipelined_inputs=["mock_input"]
+        ),  # FunctionMap response
+    ]
```
WillEngler, line 132: Interesting that we have to do this double patch - probably something we'll want to clean up

WillEngler (approved):
This looks like a great first pass! Clean and readable

garden-backend-service/tests/api/routes/modal/test_invocations.py
hholb: Yeah, I think it is because we are using the `side_effect` parameter instead of `return_value`. It threw an error on the second `POST` request without the second side effect.

PR #190, opened 2024-10-22
Include modal functions in garden search results

### LINKED ISSUE #189: Gardens created from Modal apps are immediately visible from the search page
When a user creates a garden from a Modal app, they or other users should be able to find it immediately from the search page on the frontend. 

This involves the following:
- The contents of Modal functions should be used to drive search results, in the same way that the contents of entrypoints are currently used.
- If there are any mechanisms that prevent newly created gardens from showing up in search, we should reconsider them

hholb:
Resolves: #189 

## Overview

This PR adds the necessary SQL statements to include modal functions in garden search results the same way entrypoints are considered. This adds a materialized view called `modal_functions_documents` for storing the text search documents for modal functions, as well as an index, and a trigger to update the view. The documents include the title, description, tags, and authors for each modal function. If a modal function matches the search query, its associated garden is included in the results.

## Discussion

Newly created gardens do not show on the search page. The frontend filters gardens where `doi_is_draft` is true. Maybe we want to add a `hidden` or `visibility` attribute to gardens and use that to filter gardens on the frontend separate from the doi status?

## Testing

Manual testing with my local database.

### DIFF
--- a/garden-backend-service/src/api/search/sql.sql
+++ b/garden-backend-service/src/api/search/sql.sql
@@ -1,5 +1,7 @@
-CREATE OR REPLACE FUNCTION search_gardens(search_query TEXT)
-RETURNS TABLE (garden_id int, rank real) AS $$
+CREATE OR REPLACE FUNCTION search_gardens(search_query text)
+RETURNS TABLE(garden_id int, rank real)
+AS
+    $$
 DECLARE
     query tsquery := websearch_to_tsquery(search_query);
 BEGIN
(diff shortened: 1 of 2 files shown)


garden-backend-service/src/api/search/sql.sql
```diff
@@ -37,35 +52,62 @@ CREATE MATERIALIZED VIEW IF NOT EXISTS garden_documents AS
...
+CREATE OR REPLACE FUNCTION refresh_modal_function_documents()
+RETURNS TRIGGER
+AS $$
+BEGIN
+    REFRESH MATERIALIZED VIEW modal_function_documents;
+    RETURN NEW;
+END;
+$$
+LANGUAGE plpgsql
+;
```
WillEngler, line 110: Just curious, is there some plpgsql linter running now that wasn't before?

WillEngler (approved):
This looks great, thank you Hayden!

garden-backend-service/src/api/search/sql.sql
hholb: I added a [SQL formatter](https://pypi.org/project/sql-formatter/) to my local dev setup. I'm not sold on adding it to the dev dependencies yet.

Issue #187, opened 2024-10-21
Improved error reporting from file validation, publication, and invocation endpoints


WillEngler:
This is a pretty broad ticket - it'll be up to taste for whoever picks it up to chart out the specific cases they want to hit.

## Motivation

Right now, a lot can go wrong when trying to publish a Modal function. But from the perspective of a user looking at the frontend form, it's hard to tell exactly what went wrong and what to do about it. We need to send over more informative errors to the frontend so it can tell the user what's what.

## Examples

- What are the different reasons that a file could fail the validation step? It could not be valid Python, it could be missing a variable called "app" that contains a Modal app. etc etc. These are distinct cases that a user should know about.
- What are the reasons that publication could fail? The sandboxed publication function could time out. Something in the user's container building instructions could fail, in which case we'd want to forward along the relevant log files from Modal and instruct them to double-check their work.

WillEngler:
I'd say Hayden has handled this. We can follow up with more distinct follow-up tickets if needed

PR #191, opened 2024-10-29
Setup middleware to handle Modal exceptions and improve logging

hholb:
Related: #187 -- does not resolve the ticket completely

## Overview

This PR sets us up for better error reporting to the frontend when there are issues with modal deployment/invocations that users should know about, as well as improving the error logs coming from the fastapi server.

This PR defines a new custom exception type `ModalException` that includes fields for:
- details about the error
- suggested fix for the user if it is something they can fix in their code
- http status code

There is logic added to the exception handling middleware that converts a `ModalException` to a `JSONResponse` before sending the response to the client. The middleware logs important details about the exception as well as a stack trace so we can have a better idea of how to help users when something goes wrong. A big improvement in the logging comes from a helper function to remove excess noise from the stack trace.

Since the modal validation/deployment/invocation routes are still in flux, I figured getting a scaffolding in place so that we can raise a `ModalException` with a suggested fix for the user is a good first step. This PR doesn't include errors messages for everything that could wrong, just gets a system in place so we can write specific error messages as we get stuff ironed out. I added a couple `ModalException`s into the `add_modal_app` route so we can see how this setup works in context, however I expect these errors will change as the `add_modal_app` route gets tweaked.

## Discussion

I am not sure that defining the `ModalException` in the dependencies module makes the most sense, but I haven't thought of a better place. I am open to suggestions.

## Testing

Manual testing

### DIFF
--- a/garden-backend-service/src/api/dependencies/modal.py
+++ b/garden-backend-service/src/api/dependencies/modal.py
@@ -1,5 +1,6 @@
 import modal
-from fastapi import Depends
+from fastapi import Depends, Request
+from fastapi.responses import JSONResponse
 
 from src.config import Settings, get_settings
 
@@ -7,4 +8,29 @@ from src.config import Settings, get_settings
 async def get_modal_client(settings: Settings = Depends(get_settings)) -> modal.Client:
     return await modal.client._Client.from_credentials(
         settings.MODAL_TOKEN_ID, settings.MODAL_TOKEN_SECRET
+    )
+
+
+class ModalException(Exception):
+    def __init__(
+        self,
+        detail: str,
+        status_code=400,
+        suggested_fix: str | None = None,
+    ) -> None:
+        super().__init__(detail)
+        self.detail = detail
+        self.status_code = status_code
+        self.suggested_fix = suggested_fix
+
+
+def handle_modal_exception(request: Request, error: ModalException):
+    content = {
+        "detail": error.detail,
+        "suggested_fix": error.suggested_fix,
+    }
+
+    return JSONResponse(
+        status_code=error.status_code,
+        content=content,
     )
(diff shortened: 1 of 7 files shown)


garden-backend-service/src/api/dependencies/modal.py
```diff
@@ -8,3 +9,28 @@ async def get_modal_client(settings: Settings = Depends(get_settings)) -> modal.
     return await modal.client._Client.from_credentials(
         settings.MODAL_TOKEN_ID, settings.MODAL_TOKEN_SECRET
     )
+
+
+class ModalException(Exception):
```
OwenPriceSkelly: We won't be using these with `Depends` anywhere so I don't think this is the right place for these to live -- maybe in a new `src.exceptions` module?

garden-backend-service/src/middleware/logging.py
```diff
@@ -1,11 +1,43 @@
...
 import structlog
-from fastapi import HTTPException, Request, Response
+from fastapi import Request, Response
 from fastapi.responses import JSONResponse
 from starlette.middleware.base import BaseHTTPMiddleware
 
+from src.api.dependencies.modal import ModalException, handle_modal_exception
+
+
+def filter_stacktrace(exc: Exception, func_to_stop_at: str = "run_endpoint_function"):
```
OwenPriceSkelly: nit: can we name this helper function something like `format_traceback` or `process_traceback` to indicate that it's building a nicely cleaned up string for us and not a modified traceback object or list of `FrameSummary` objects or anything

OwenPriceSkelly: I also think we might want this helper to live in `src.exceptions`

OwenPriceSkelly (approved):
LGTM! couple superficial nits about where the new code should live but otherwise this looks great 🙌

garden-backend-service/src/api/dependencies/modal.py
hholb: Good idea! New module incoming!

Issue #185, opened 2024-10-17
Publish Modal functions from a Lambda


WillEngler:
For mysterious reasons, publishing Modal functions from Modal functions fails. (The publication goes through but the published app doesn't work.) For this ticket:

- Move sandboxed Modal operations (validation and publication) to Lambdas
- Further harden the publication function against token exfiltration

PR #192, opened 2024-10-30
Add terraform and GH actions for deploying sandboxed functions as Lambda

WillEngler:
First step of #185 

## Overview

This PR adds the Terraform, GitHub Action, and skeleton of the Lambda handler needed to run our sandboxed operations from Lambda instead of Modal. 

## Discussion

To keep it reasonably sized, this PR does not yet actually incorporate the Lambda into the API routes that perform the sandboxed operations. That will come next. This PR is focused on getting the infra and plumbing up. The next one will get it all working. (It will also deduplicate the two `sandboxed_functions` directories!)

## Testing

I tested the lambda function for Modal deployment via a Lambda I made from the dashboard. The critical thing to figure out here was how to correctly bundle up the dependencies. Unlike the old webserver Lambda we deployed, with this one I couldn't just pull binaries from pip for the appropriate platform, so it took a lot of trial and error. Shoutout to the [act](https://github.com/nektos/act) utility which sped up my iteration/testing cycle a lot.

### DIFF
--- a/infra/modules/lambda/outputs.tf
+++ b/infra/modules/lambda/outputs.tf
@@ -1,0 +1,20 @@
+output "garden_app_invoke_arn" {
+  value = aws_lambda_function.sandboxed_app.invoke_arn
+}
+
+output "garden_app_function_name" {
+  value = aws_lambda_function.sandboxed_app.function_name
+}
+
+output "garden_authorizer_invoke_arn" {
+  value = aws_lambda_function.sandboxed_app.invoke_arn
+}
+
+output "garden_authorizer_function_name" {
+  value = aws_lambda_function.sandboxed_app.function_name
+}
+
+
+output "lambda_exec_role_name" {
+  value = aws_iam_role.lambda_exec.name
+}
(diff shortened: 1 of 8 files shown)


infra/modules/lambda/outputs.tf
```diff
@@ -0,0 +1,20 @@
+output "garden_app_invoke_arn" {
+  value = aws_lambda_function.sandboxed_app.invoke_arn
+}
+
+output "garden_app_function_name" {
+  value = aws_lambda_function.sandboxed_app.function_name
+}
+
+output "garden_authorizer_invoke_arn" {
```
OwenPriceSkelly: is this vestigial from old code or will we need to have an authorizer lambda too?

OwenPriceSkelly (approved):
LGTM

infra/modules/lambda/outputs.tf
WillEngler: Vestigial, thanks for the catch

PR #194, opened 2024-11-11
Stop runaway modal functions synchronously

### LINKED ISSUE #193: Prevent runaway Modal functions
(First part of async epic)

This ticket will lay the groundwork for letting users perform a model invocation via Modal that takes longer than the gateway timeout limit for the FastAPI app.

This ticket will not yet make any API changes visible to clients of the backend. It will just move Modal invocations to a background task and apply a maximum execution time to it. The background task will update a new table with the status of the invocation (pending vs error vs done) and store the result of the invocation. The API route will block and poll the DB table for the result - logic that will eventually be moved out to the clients.

Success criteria:
- Trigger a function call that would normally cause a crash loop with Modal. With this new functionality in place, it should end after the time limit that we set.

hholb:
Resolves: #193 

## Overview

This PR sets up the modal invocations route to cancel invocations if they run longer than a new `MODAL_TIMEOUT_SECONDS` setting. Modal invocations are still synchronous in this PR, but the machinery is largely in place to move to asynchronous invocations in the future. 

Logic for retrieving the outputs, or cancelling the invocation, from a modal function is now handled in a helper function called `monitor_modal_invocation` that can be used in a background task. Currently we await the output/timeout synchronously so clients are not effected by this PR.

Asynchronous invocations will come in a later PR.

The for estimating usage per invocation has been moved into the helper functions as well. This will be handled by the background task in the future.

## Testing

Lots of manual testing using personal modal account to verify that timed out invocations are correctly cancelled.
Unit tests will come in another PR.


garden-backend-service/src/api/routes/modal/invocations.py
```diff
@@ -68,43 +68,32 @@ async def invoke_modal_fn(
...
+        function_call_id=invocation.function_call_id,
     )
+    db.add(modal_invocation)
     await db.commit()
 
-    if not outputs_response.outputs:
-        log.warning("No outputs received from function call")
-        raise Exception("No outputs received from function call")
+    # Until we have a route for polling a background task, just await the result directly
+    await monitor_modal_invocation(invocation, modal_invocation, modal_client, settings)
```
WillEngler: I don't really know where I'd take it, but fyi that I found the variable names "invocation" vs "modal_invocation" confusing. I'd expect `modal_invocation` to be more Modal-y, but in fact it's our DB representation of an invocation, whereas `invocation` is actually an object from the Modal SDK.

Not a big deal, but noticed that it was tripping me up and wanted to pass that along.

garden-backend-service/src/modal/utils.py
```diff
@@ -0,0 +1,81 @@
...
+    client: modal.Client,
+    settings: Settings,
+):
+    session_maker = await get_db_session_maker(settings=settings)
+    async with session_maker() as session:
+        if inv := await ModalInvocation.get(session, id=db_invocation.id):
+            try:
+                # Try and get the invocation outputs
+                outputs_response = await invocation.pop_function_call_outputs(
+                    timeout=settings.MODAL_TIMEOUT_SECONDS,
```
WillEngler, line 54: I don't follow what's happening here. If MODAL_TIMEOUT_SECONDS is 5 seconds by default, does the Modal function have a maximum of 5 seconds to run?

WillEngler (commented):
I don't think I understand the main thrust of this yet. (The 5 second default is throwing me.) Probably something we can clear up on a call.

WillEngler (approved):
I think this is good, we'll just want to bump MODAL_TIMEOUT_SECONDS considerably

garden-backend-service/src/api/routes/modal/invocations.py
hholb: Thank you for the feedback! I will try to make the names more clear.

garden-backend-service/src/modal/utils.py
hholb: `pop_function_call_outputs` ([src](https://github.com/modal-labs/modal-client/blob/2aedf2b1e40a61ffa09ae2392bd7c62087e498d8/modal/functions.py#L148)) will poll Modal's backend until the invocation outputs are available or the  `timeout` is reached. If the timeout is reached before the invocation has completed, `pop_function_call_outputs` returns a response object with a non-zero value in `num_unfinished_inputs` [(src).](https://github.com/modal-labs/modal-client/blob/2aedf2b1e40a61ffa09ae2392bd7c62087e498d8/modal_proto/api.proto#L1442) To cancel the invocation and shutdown the containers we can send a `FuncionCallCancelRequest` [(src).](https://github.com/modal-labs/modal-client/blob/2aedf2b1e40a61ffa09ae2392bd7c62087e498d8/modal_proto/api.proto#L1255)


I bumped the default for `MODAL_TIMEOUT_SECONDS` to 300 seconds, does that seem like decent default?

WillEngler: Thank you for spelling that out! I think 300 seconds is a decent place to start once the async routes are in place. For the purpose of this PR, maybe 20 or 25 seconds makes sense to come in under the gateway timeout that is currently in place on the FastAPI app (which I think is 30 seconds)

PR #195, opened 2024-11-11
Use newly deployed Lambdas to run sandboxed operations

WillEngler:
Implements #185 

## Overview

This PR makes the publication route (POST /modal-apps) use Lambda instead of Modal for remote sandboxed operations. 

## Discussion

The sandboxed_functions.py file feels awkward and boiler platey, but at least right now I think it's the least bad way to go. It abstracts away the difference between remote and local execution so that you should (hopefully) just be able to look at the lambda functions and the route code to understand how things fit together.

## Testing

I tested this manually. I was able to run POST /modal-apps (which exercises both remote functions) both locally and as deployed in Lambda. I also executed the Modal function I deployed that way remotely, confirming that this resolves the crash loop issue we saw when trying to deploy via Modal.


garden-backend-service/src/api/dependencies/sandboxed_functions.py
```diff
@@ -1,47 +1,105 @@
...
+    elif "Exception" in response_payload:
+        details = response_payload["Exception"]
+        raise Exception(
+            detail=details["detail"],
+        )
+
+
+def make_lambda_invoker(
+    function_name: str,
+    sub_function_name: Literal["deploy_modal_app", "validate_modal_file"],
```
OwenPriceSkelly, line 47: does it simplify things for this helper if we're not planning to validate modal files in a lambda?

garden-backend-service/src/api/dependencies/sandboxed_functions.py
```diff
@@ -1,47 +1,105 @@
...
-from src.sandboxed_functions.modal_publishing_helpers import (
+from src.exceptions.modal import ModalException
+from src.sandboxed_functions.lambda_function import (
+    DeployModalAppArgs,
+    ValidateModalFileArgs,
     deploy_modal_app,
     validate_modal_file,
 )
 
+# This module handles boilerplate for turning our sandboxed code execution functions into FastAPI dependencies.
```
OwenPriceSkelly, line 16: slightly more pythonic I think to have this block comment as the module docstring (if the first statement in a module is a str literal it becomes the module's `__doc__`)

garden-backend-service/src/api/dependencies/sandboxed_functions.py
OwenPriceSkelly: (note: we synced and decided to leave as-is)

OwenPriceSkelly (approved):
LGTM!

garden-backend-service/src/api/dependencies/sandboxed_functions.py
WillEngler: Ooh good call

PR #200, opened 2024-11-12
Create routes for async modal invocation flow

### LINKED ISSUE #196: Routes for handling Modal invocations asynchronously
Part of the async Modal epic. Related: #194 

To handle long-running Modal invocations asynchronously , we need a couple of new routes:
- POST route for initiating the invocation which then returns an invocation ID to the client while the function executes.
- GET route where the client can poll for the status/outputs of an invocation

hholb:
Builds on: #194 
Resolves: #196 

## Overview

This PR adds a POST route that mirrors the current `invoke_modal_fn` route, but returns a invocation ID to the client and handles the modal invocation in a background task. There is a new GET route that takes an invocation ID and returns the status of the invocation, any error messages, or outputs if the invocation has completed.

## Discussion

I created a new `POST` route for requesting an invocation via the async flow so that we can keep the synchronous route up until we decide we are ready to get rid of it.

The logic is pretty straightforward here, the only tricky part was getting the response type of the GET route correct. I kept the `_ModalGenericResult` as a field of the new `ModalInvocationOutputsResponse` that also has fields for the status of the invocation and an optional field for error info. This should allow the SDK to grab the `_ModalGenericResult` from the response and pass it into the Modal SDK and let Modal handle parsing the outputs.

## Testing

A couple of new unit test of the new routes.


WillEngler (approved):
LGTM! 🎂

Issue #198, opened 2024-11-12
Support Modal classes (not just functions)


WillEngler:
Modal classes are very useful. They let users bake model weights into their containers, among many other benefits. We should support them at publication and invocation time.

PR #201, opened 2024-11-18
Publish and invoke methods on Modal classes

WillEngler:
Backend portion of #198

## Overview

This PR tweaks the Modal backend routes so that
- You can publish a Modal app that has a class
- You can invoke class methods

## Discussion

I tweaked the POST /modal-apps logic so that the functions/methods a user submits just need to be a subset of the functions/methods detected in the submitted app. On a Modal class, there is always a "ClassName.*" method that we should not persist in the DB. This flexibility will also let users skip registering @modal.build and @modal.enter methods.

Also, I hit the same security flag on starlette we saw in the SDK. Seems like it could more plausibly affect the API, so I bumped fastapi and starlette

## Testing

I added a new publication test case with a class. I tested the flow manually by creating and invoking a method with the synchronous routes.


OwenPriceSkelly (approved):
LGTM!

PR #202, opened 2024-11-19
modal app validation

### LINKED ISSUE #162: POST /modal-file-validation
We need an endpoint that validates Modal files and extracts metadata from them if they are valid. This will be used to enhance the frontend modal app submission page.

Metadata to return will include:
- app name
- function names
- base image and pip requirements
- individual function text

This entrypoint (and the helper function that drives it) should also reject Modal apps that do things Garden doesn't want them to do. eg, setting up cron timed execution should be rejected

OwenPriceSkelly:
closes #162 

## Overview

This PR fleshes out the validation process for user modal files. The bulk of the PR is in a new module for parsing modal files to extract app/function metadata. 

We use this module two places. First, we use it in a new route (I went with `POST /modal-file-metadata` instead of `validation` but very open to pushback) which the frontend can use to prefill the upload form for the user. This route takes a modal file and returns everything you need for the subsequent `POST /modal-apps` request which then actually creates and deploys the app. Among other things, if the ast scan sees a function argument or image method that we don't want to allow, it will raise a helpful `ModalException` 

The `POST /modal-apps` route has also been updated to use the new module in order to verify that the metadata in the request body matches the uploaded file contents. 

The full file upload/validation flow would look like this: 
  - frontend sends user's file contents to `POST /modal-file-metadata` and gets the app/function metadata in response (or a helpful `ModalException` is raised)
  - frontend uses that metadata to pre-fill the form for `POST /modal-apps`
  - User adds/edits/confirms the form metadata, which is sent to `POST /modal-apps` to double-check and finally deploy the app. 

## Discussion

Once I came to the conclusion that we couldn't extract image info by exec'ing user code, I was pretty much all-in on the ast approach. Unfortunately, modal's flexibility for specifying a gpu/cpu spec made it impractical to get that particular bit of metadata from just the ast. 

So I'm left in a vaguely unpleasant compromise where we get 99% of the app/function metadata from the ast, but we still need the sandboxed lambda function to get the `hardware_spec`s. I can live with it. 

## Testing

leaving PR as a draft until I've finished with tests


garden-backend-service/src/api/routes/modal/modal_apps.py
```diff
@@ -82,14 +68,86 @@ async def add_modal_app(
...
+        raise ModalException(
+            detail=f"Base image mismatch: Got '{app_info.image.base_image}' from file but '{app_metadata.base_image_name}' in request.",
+            suggested_fix="Ensure the base image in your modal file matches the metadata.",
+        )
+
+    # Check requirements match
+    request_pip_requirements = app_metadata.requirements
+    file_pip_requirements = app_info.image.pip_requirements
+    if set(request_pip_requirements) != set(file_pip_requirements):
+        logger.warning(
```
WillEngler: Curious why we're logging this. Do we care to track this information as admins? Or is this something we should pass on to the user?

garden-backend-service/src/api/routes/modal/modal_apps.py
```diff
@@ -82,14 +68,86 @@ async def add_modal_app(
...
+            file_pip_requirements=file_pip_requirements,
+        )
+
+    request_conda_requirements = app_metadata.conda_requirements
+    file_conda_requirements = app_info.image.conda_requirements
+    if set(request_conda_requirements) != set(file_conda_requirements):
+        logger.warning(
+            "Requirements mismatch between file and metadata.",
+            request_conda_requirements=request_conda_requirements,
+            file_conda_requirements=file_conda_requirements,
```
WillEngler: Same here

garden-backend-service/src/api/routes/modal/modal_apps.py
```diff
@@ -82,14 +68,86 @@ async def add_modal_app(
...
+    # confirm function names match
+    request_function_names = {fn.function_name for fn in app_metadata.modal_functions}
+    file_function_names = {fn.function_name for fn in parse_results.functions}
+
+    if not request_function_names:
+        raise ModalException(
+            detail="No function names provided",
+            suggested_fix="Must provide names of functions to expose in request to create Modal App",
+        )
+    if not request_function_names <= file_function_names:
```
WillEngler, line 228: TIL about this set operation notation

garden-backend-service/src/api/schemas/modal/modal_app.py
```diff
@@ -29,3 +35,13 @@ class ModalAppMetadataResponse(ModalAppMetadata):
     @property
     def modal_function_ids(self) -> list[str]:
         return [mf.id for mf in self.modal_functions]
+
+
+class ModalParseMetadataRequest(BaseSchema):
+    file_contents: str
+
+
+class ModalParseMetadataResponse(ModalAppCreateRequest):
```
WillEngler: This is a little confusing to me. Could it inherit from "ModalAppMetadata" instead? I think that would get the point across better. It's surprising that this needs to return the owner identity id.

WillEngler (approved):
Some nits, but overall I think this is looking good

garden-backend-service/src/api/routes/modal/modal_apps.py
OwenPriceSkelly: tbh just wan't thinking too hard and demoted them from an exception to a logger warning. imo not worth passing on to the user, since they would have needed to actively change those from the pre-filled form values

garden-backend-service/src/api/schemas/modal/modal_app.py
OwenPriceSkelly: yeah that's just as well, no strong preference here

PR #203, opened 2024-11-21
Add routes for async modal app publication

### LINKED ISSUE #197: Async Modal app publication
Hopefully reusing some of the infrastructure used to make POST /modal-invocations async, we'll need to do the same for POST /modal-apps. (Assuming many useful apps will take longer than 30 seconds to build.)

hholb:
Resolves: #197 

## Overview

This PR adds a new POST route at `/modal-app/async` that handles modal app deployments in a background task. The route returns metadata about the app immediately, but with a new field `deploy_status` that is the same enum type from the async modal invocations. Clients can deploy using the new async POST route and then check the status of the deployment by hitting the GET `/modals-apps/{id}` route which now includes the `deploy_status` field in the metadata response. The metadata includes an optional field called `deploy_error` that can hold a string describing any errors that are thrown during the deployment.

## Discussion

The background task that monitors the deployment does not have any timeout logic. It seems like that might be better handled in the lambda?

## Testing

Added new unit tests


garden-backend-service/src/modal/utils.py
```diff
@@ -76,6 +81,28 @@ async def monitor_modal_invocation(
...
+):
+    session_maker = await get_db_session_maker(settings=settings)
+
+    try:
+        deploy_func(deploy_config)
+        async with session_maker() as session:
+            if modal_app := await ModalApp.get(session, id=db_modal_app.id):
+                modal_app.deploy_status = AsyncModalJobStatus.DONE
+                await session.commit()
+    except Exception as e:
```
OwenPriceSkelly, line 103: understanding check: do we not want to re-raise this exception because it's problematic to raise an exception from a background task?

OwenPriceSkelly: if so I think it's a good idea to log some information about the exception (since it won't get picked up by the exception logging middleware)

OwenPriceSkelly (approved):
LGTM! curious about exception handling but nothing to block over 🌱🚢

PR #204, opened 2024-11-21
Add gate on Modal Publishers group membership

WillEngler:
Closes #188 

## Overview

This PR adds a check for membership in the Modal publishers Globus Group to gate access to the publication routes.

## Discussion

This supersedes the previous hardcoded "Modal VIP List". It also narrows the gating down to the risky routes - the ones that are exec'ing user code.

## Testing

I tested manually on a protected route. (Confirming with one user identity in the group and another identity not in the group.) I also tweaked the affected unit tests to accommodate the auth changes.

PR #205, opened 2024-11-25
hotfix: update modal publishing routes to use new group dependency, update CI

hholb:
## Overview

After merging #204, I merged #203. There were no conflicts and the merge went fine -- or so I thought. I realized there were some changes from #204 that didn't make it into #203 that are causing some import errors. Specifically, I didn't catch the new `in_modal_publishers_group` dependency leading to import errors trying to import the old `modal_vip` dependency. This PR remedies the issue.

From this, I realized our CI workflow for running the test suite was not configured to rerun the tests when merging a PR into dev. This PR sets up the action to trigger on merges to main and dev, hopefully this will catch this kind of error in the future.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,10 @@ on:
   pull_request:
     branches:
       - "*"
+  push:
+    branches:
+      - "dev"
+      - "main"
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


OwenPriceSkelly (approved):
good catch! lgtm

PR #206, opened 2024-11-26
Define modal app overwriting behavior

### LINKED ISSUE #199: Define Modal app overwriting behavior
What happens when users upload a Modal app with name "foo", and then upload another Modal app with name "foo"? What should happen? This either causes an error or allows an existing Modal app to be quietly overwritten. Whatever the right behavior is, we need to make it unsurprising and obvious to the user.

hholb:
Resolves: #199 

## Overview

This PR defines our modal app overwriting behavior. By default, we follow Modal's lead to overwrite apps with the same name. This is controlled by a new parameter on the POST request called `overwrite_existing` which defaults to `true`. If a request comes in where this field is `false` and the app name conflicts with an existing app, the route returns a 409 error.

## Discussion

We can easily change this behavior by changing the default for the `overwrite_existing` field of the create request. 

## Testing

New tests!

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,6 +3,10 @@ on:
   pull_request:
     branches:
       - "*"
+  push:
+    branches:
+      - "dev"
+      - "main"
   workflow_dispatch:
 permissions:
   id-token: write # This is required for requesting the JWT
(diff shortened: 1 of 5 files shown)


WillEngler (approved):
LGTM! I like the overwrite_existing option being there - that gives us the option to "fail by default" on the frontend which is probably what we want

PR #214, opened 2024-12-11
Add is_test bool to gardens

WillEngler:
Yak shaving for https://github.com/Garden-AI/garden-frontend/issues/228

## Overview

This PR adds a new is_test field to gardens. This will let us save whether users would like their gardens to show up in search without relying on the draft/non-draft state of the garden's DOI. The migration initializes this field for old gardens with the value of doi_is_draft

## Discussion

In general I want to move us away from linking gardens' state to their DOI. I want to let people's gardens show up in search without them committing to publishing a DOI. (think about our upcoming hackathons for example)

## Testing

I made a Garden with the old DB model and then applied the migration to make sure the migration will work. I also updated old garden test fixtures with the new field


OwenPriceSkelly (approved):
LGTM! 🤠

PR #217, opened 2025-01-06
handle large args/results

### LINKED ISSUE #164: feat: POST /modal-invocations can handle large arguments/results
If serialized arguments/results from modal are >2MiB, the modal client will switch to a blob storage upload/download instead of sending the raw bytes directly. 

This is likely to be more involved on the sdk-side than the backend, but the `FunctionInput` payload that `_create_invocation` builds will need to be able to specify a `args_blob_id` instead of always sending raw bytes as `args`. The response schema shouldn't need to be changed.

OwenPriceSkelly:
fixes #164 (Draft PR until tests are finished)

## Overview

This supports large input args *and* large results from modal functions. There will be a corresponding sdk PR to use the new blob upload url route. 

## Discussion

Any additional context that you think is important for reviewers or future users/contributors looking at this PR.

## Testing

How did you test this? If a reviewer pulled this branch down and wanted to kick the tires, how would they do so?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,3 +176,4 @@ response.json
 .idea/
 *.zip
 /garden-backend-service/repopack-output.txt
+/garden-backend-service/.basedpyright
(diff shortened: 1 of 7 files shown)


WillEngler (approved):
LGTM!

PR #219, opened 2025-01-07
Another pass at internal modal app name generation

### LINKED ISSUE #218: Generated modal app names exceed size limit
When publishing a modal app with this name: `app = modal.App("iris-classifier")` I got this deployment error:
```
Invalid App name: 'e9a17e09-657b-4087-a719-241ab72b1d9b-iris-classifier-32949b7e-bbc4-4b5c-8f81-511b28a0e8f4'.

Names may contain only alphanumeric characters, dashes, periods, and underscores, must be shorter than 64 characters, and cannot conflict with App ID strings.
```

hholb:
Resolves: #218 

## Overview

This PR fixes a small issue I ran into where our generated modal app names hit modal's size limit of 64 characters.

When publishing a modal app `app = modal.App("iris-classifier")` I got this deployment error:
```
Invalid App name: 'e9a17e09-657b-4087-a719-241ab72b1d9b-iris-classifier-32949b7e-bbc4-4b5c-8f81-511b28a0e8f4'.

Names may contain only alphanumeric characters, dashes, periods, and underscores, must be shorter than 64 characters, and cannot conflict with App ID strings.
```

## Discussion

Updated the logic in the `_generate_app_name` helper to enforce the app name size constraint transparently from the users perspective. Our users can name their app anything that is valid on Modal. If the name is too large after we add the prefix and unique suffix, the user provided app name is truncated to fit within the size limit. This gives us the unique names we need internally, but doesn't affect the user experience.

## Testing

Added a new unit test to verify this behavior

### DIFF
--- a/garden-backend-service/src/api/routes/modal/modal_apps.py
+++ b/garden-backend-service/src/api/routes/modal/modal_apps.py
@@ -227,8 +227,31 @@ def _validate_modal_app_metadata_helper(
 
 
 def _generate_app_name(user: User, app_name: str) -> str:
+    """Generate a unique app name for deployment to Modal.
+
+    The app names look like: <user_id>-<app_name>-<unique-suffix>
+
+    Note: Modal limits app names to 64 characters. To give our users
+    the freedom to name their apps within Modal's guidelines, this function
+    will truncate the user supplied app name to fit within the size limit before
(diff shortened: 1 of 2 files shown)


WillEngler (approved):
Good catch, thank you!!

PR #220, opened 2025-01-09
Catch invalid modal app names on initial file validation

hholb:
Related: #187 

## Overview

This RP adds logic to catch invalid Modal App variable names on initial validation. I was able to get past the intial validation with a modal app variable named: `my_app = modal.App()`. On final deployment, the lambda that deploys to modal to failed saying it could not find a identifier named 'app'. The solution looks through the app identifiers we pull out during initial validation and raises an error if none of them are named `app`.

## Testing

Manually using a local fronted


WillEngler (approved):
LGTM! This is an important catch

garden-backend-service/src/modal/user_file_parsing.py
```diff
@@ -124,6 +124,12 @@ def parse_modal_file(contents: str) -> ModalFileParseResults:
                 for function_info in class_function_infos:
                     functions[function_info.function_name] = function_info
 
+    if not any(name == "app" for name in apps.keys()):
```
OwenPriceSkelly:
```suggestion
    if "app" not in apps:
```

OwenPriceSkelly (approved):
LGTM! ✨ pythonic ✨ style nit but nothing to block over

PR #232, opened 2025-01-29
feat: local_entrypoints populate `test_functions` for functions they call 

### LINKED ISSUE #227: feat: ModalFunctions have test_functions
As a garden user who has discovered a ModalFunction, I want see an example invocation of the function of interest (like regular entrypoints). 

Entrypoints have a test_functions attribute, which is populated with the source of any `@entrypoint_test`-decorated functions found in the notebook. The closest analogue in modal-world is the `@local_entrypoint` decorator (no relation), which may invoke modal functions like `myfunc.remote(<sample input>)`. 

A first pass at this could just extract any local_entrypoint(s) source and store it in test_functions field, which is probably enough for most of users.

OwenPriceSkelly:
closes #227 

## Overview
Any `@local_entrypoint`s that appear to exercise a user's modal function in their uploaded file are now persisted as `ModalFunction.test_functions` so the frontend can showcase example usage. 

## Discussion

This is the first pass to just extract and persist local_entrypoint source code for modal function metadata. Future passes could implement the logic to bestow them with `example_usage` field that looks like what a garden user's typical invocation would, rather than just the local_entrypoint text.  

## Testing

extended the unit test for class-based functions, plus added a test with an edge case for multiple local_entrypoints that call distinct functions

### DIFF
--- a/garden-backend-service/tests/fixtures/modal_file_constants.py
+++ b/garden-backend-service/tests/fixtures/modal_file_constants.py
@@ -98,8 +98,39 @@ class Model:
     @modal.method()
     def predict_proba(self, x):
         return self.model.predict_proba(x)
+
+@app.local_entrypoint()
+def main():
+    Model.predict.remote("abc")
+    Model.predict_proba.remote("123")
 """
 
+# Valid case: Different local entrypoints for different functions
+VALID_MULTIPLE_LOCAL_ENTRYPOINTS = """
+import modal
+
+app = modal.App("science-app")
+
+@app.function()
+def science_stuff():
+    import numpy as np
+    return np.random.rand(10)
+
+@app.local_entrypoint()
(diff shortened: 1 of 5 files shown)


hholb (approved):
LGTM!

garden-backend-service/tests/fixtures/modal_file_constants.py
```diff
@@ -98,8 +98,39 @@ def predict(self, x):
...
+
+@app.function()
+def ml_stuff():
+    import torch
+    return torch.rand(10)
+
+
+@app.local_entrypoint()
+def test_ml_stuff():
+    results = ml_stuff.remote("BILL! BILL! BILL! BILL!")
```
hholb, line 131: :notes:

PR #233, opened 2025-01-29
Quick and dirty load tests

### LINKED ISSUE #229: Quick and dirty backend load test
I made some quick changes to the backend deployment ahead of the ESIP event. (Moved from one uvicorn worker to 5 and bumped the instance size.) It would be great to gather some quick empirical evidence on how this works to confirm I set it up correctly. Something like an [Artillery](https://github.com/artilleryio/artillery) test that simulates 10 concurrent users uploading and executing models. Can we actually handle a workshop of 10 people really exercising the service at the same time?

hholb:
Resolves: #229 

## Overview

This PR adds load tests of the async modal publishing, and async invocation flows.

Included are fixes with our async setup that I discovered along the way including:

- a race condition when getting concurrent requests with a new user that caused db integrity errors
    - fixed by handling user creation logic in the authed_user dependency
- an issue in the async POST route with how we were passing the `db_modal_app` to the background task that tied it to the db session of the request.
    - fixed by passing the id of the new modal app to the background task and having the background task query for the model using a new session

## Discussion

We hit an error when sending >~30 concurrent deployment requests that comes from asyncpg hitting its connection pool limit. Setting the engine pool size parameter to a higher value , e.g. `create_async_engine(..., pool_size=50)`, doesn't seem to have any effect on the number of concurrent requests we can serve with our current setup. I will keep thinking about this, but we can service a decent amount of requests right now so we can kick the can down the road a bit here.

## Testing
I added a CLI arg for pytest `--num-concurrent-requests` (defaults to 30) so we have any easy way of testing different sized loads.

```shell
uv run pytest --num-concurrent-requests 50 tests/load_test.py
```

The number of concurrent requests is configurable via an environment variable `GARDEN_TEST_NUM_CONCURRENT_REQUESTS`:

```shell
GARDEN_TEST_NUM_CONCURRENT_REQUESTS=50 uv run pytest tests/load_test.py
```

Pytest prioritizes the cli arg, for example:

```shell
# will run with 5 concurrent requests
GARDEN_TEST_NUM_CONCURRENT_REQUESTS=50 uv run pytest --num-concurrent-requests 5 tests/load_test.py
```

### DIFF
--- a/garden-backend-service/src/models/base.py
+++ b/garden-backend-service/src/models/base.py
@@ -1,6 +1,6 @@
 from typing import Any, Optional, Type, TypeVar
 
-from sqlalchemy import exc, select
+from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncAttrs, AsyncSession
 from sqlalchemy.orm import DeclarativeBase
 from sqlalchemy.orm.attributes import QueryableAttribute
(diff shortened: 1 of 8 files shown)


garden-backend-service/src/models/base.py
```diff
@@ -29,16 +29,27 @@ async def get(
     async def get_or_create(
```
OwenPriceSkelly: Hot take: since we only actually use the `get_or_create` helper in the authed_user dependency, do we really need it our base object? This method is starting to feel specific to the authed_user usage of it (for example, is it always unlikely that the object might have disappeared? or is that only a reasonable assumption for now because we never delete from the users table?), and I wonder if it might be easier to reason about if we just pulled the integrity handling logic out to the dependency itself and cut our losses on the helper method.

I also am not totally seeing how this manages to avoid the race condition, though I'm not 100% sure I see where the race condition came from in the first place. I think I'm missing some ORM subtlety most likely

garden-backend-service/tests/conftest.py
```diff
@@ -398,3 +401,76 @@ def mock_is_doi_registered(mocker):
...
+                ),
+                data_format=api_pb2.DATA_FORMAT_PICKLE,
+            )
+        ]
+    )
+    return mock_invocation
+
+
+@pytest.fixture
+def modal_test_environment(
```
OwenPriceSkelly, line 436: this seems like it could be very handy

OwenPriceSkelly (approved):
LGTM! 

the new fixtures/tests seem well thought out (as usual, glory to the test czar 🫡), and nice catch using an ID instead of an object for the background task. 

let's sync up (hah) on the race condition bit which I'm not sure I totally grok, but otherwise no gripes 🙌

PR #234, opened 2025-01-30
Fix logging issue caused by startlette middleware base class

### LINKED ISSUE #228: Restore custom logging
I had to disable our custom logging middleware because of a buggy starlette-fastAPI interaction. We would like to have the ability to have logging middleware and get these additions back.

hholb:
Resolves: #228

## Overview

This PR fixes the issue we were having with background tasks and the starlette middle ware base classes.

Using this approach: https://fastapi.tiangolo.com/tutorial/middleware/ appears to get around the issue with little fuss.

## Testing

Tested manually by deploying a modal app with function that sleeps for 10 seconds, invoked the function via the SDK, and watched the logs. The logs showed a response immediately after the request was initiated, as well as the logs for the repeated GET requests polling the invocation status before returning the result.

### DIFF
--- a/garden-backend-service/src/main.py
+++ b/garden-backend-service/src/main.py
@@ -22,9 +22,9 @@ from src.api.routes import (
 from src.api.routes.mdf import search as mdf_search
 from src.config import Settings, get_settings
 from src.middleware.logging import (
-    ErrorHandlingMiddleware,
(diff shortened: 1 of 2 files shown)


WillEngler (approved):
Excellent!

PR #235, opened 2025-02-03
artillery load test, singleton engine object

hholb:
Related: #229 

## Overview

This PR adds a simple load test using [artillery](https://www.artillery.io/docs) to gather some metrics on how many concurrent requests we can handle and discover issues with our async setup.

While running the load tests I discovered an issue with how we were using the sqlalchemy engine object in our routes. This PR includes a fix that increases the number of concurrent requests we can handle by a significant amount, see metrics below.

### The load test

The load test hits a couple of new routes defined in `garden-backend-service/api/routes/load-test.py`.
The load test routes are only added to the app if the `GARDEN_ENV`  is `local`, or `dev` This allows us to run the load test against local containers and our dev deployment, but does not expose these routes in prod.

The test is a simplified route that mimics our background task setup for async modal routes, but allows the test to run without need to auth. Basically, we send a request to the `load-test/` route, the route starts a background task and immediately returns a successful status. The background task runs a couple of database queries and sleeps for a bit to mimic a long running background task. The test ramps up to sending 50 requests per second sustained for 30 seconds.

The Artillery setup is located in `load-tests/` and the test can be initated by cd-ing into the `load-tests` directory and running:
```
npm run test:local # runs against http://localhost:5500
npm run test:dev   # runs against https://api-dev.thegardens.ai
```

#### Baseline

Running the load test against our current dev branch quickly exposed a problem:

```
All VUs finished. Total time: 1 minute, 44 seconds

--------------------------------
Summary report @ 12:40:36(-0500)
--------------------------------

errors.ETIMEDOUT: .............................................................. 2553
http.codes.200: ................................................................ 175
http.downloaded_bytes: ......................................................... 2580
http.request_rate: ............................................................. 30/sec
http.requests: ................................................................. 2725
http.response_time:
  min: ......................................................................... 8
  max: ......................................................................... 9903
  mean: ........................................................................ 3190.5
  median: ...................................................................... 2143.5
  p95: ......................................................................... 9230.4
  p99: ......................................................................... 9801.2
http.response_time.2xx:
  min: ......................................................................... 8
  max: ......................................................................... 9903
  mean: ........................................................................ 3190.5
  median: ...................................................................... 2143.5
  p95: ......................................................................... 9230.4
  p99: ......................................................................... 9801.2
http.responses: ................................................................ 175
plugins.expect.ok: ............................................................. 172
plugins.expect.ok.statusCode: .................................................. 172
vusers.completed: .............................................................. 172
vusers.created: ................................................................ 2725
vusers.created_by_name.Load Test: .............................................. 2725
vusers.failed: ................................................................. 2553
vusers.session_length:
  min: ......................................................................... 16.4
  max: ......................................................................... 10001.9
  mean: ........................................................................ 3274.6
  median: ...................................................................... 2276.1
  p95: ......................................................................... 9230.4
  p99: ......................................................................... 9416.8
```

Notice that out of 2725 requests, only 175 were successful. An audit of the container logs during the test showed we were hitting postgres' client limit when servicing a low number of requests. This points to a problem with our database connection setup. Manual testing showed that we hit the limit ~30 concurrent requests.

#### After singleton engine

The issue  that caused an error around 30 concurrent requests was due to creating a new engine object for each request. The engine maintains a connection pool to the database which means when servicing a handful of requests concurrently, each request creates its own pool which quickly exhausted postgres' connection limit. The solution is a singleton engine object that lives for the lifetime of the app. This is handled by a new class that enforces the singleton pattern. Check it out in the database dependency.

Another run of the load test confirms that this was the issue:

```
--------------------------------
Summary report @ 14:16:10(-0500)
--------------------------------

http.codes.200: ................................................................ 1525
http.downloaded_bytes: ......................................................... 22875
http.request_rate: ............................................................. 33/sec
http.requests: ................................................................. 1525
http.response_time:
  min: ......................................................................... 4
  max: ......................................................................... 374
  mean: ........................................................................ 19
  median: ...................................................................... 7.9
  p95: ......................................................................... 50.9
  p99: ......................................................................... 278.7
http.response_time.2xx:
  min: ......................................................................... 4
  max: ......................................................................... 374
  mean: ........................................................................ 19
  median: ...................................................................... 7.9
  p95: ......................................................................... 50.9
  p99: ......................................................................... 278.7
http.responses: ................................................................ 1525
plugins.expect.ok: ............................................................. 1525
plugins.expect.ok.statusCode: .................................................. 1525
vusers.completed: .............................................................. 1525
vusers.created: ................................................................ 1525
vusers.created_by_name.Load Test: .............................................. 1525
vusers.failed: ................................................................. 0
vusers.session_length:
  min: ......................................................................... 7.6
  max: ......................................................................... 386.3
  mean: ........................................................................ 25.8
  median: ...................................................................... 13.9
  p95: ......................................................................... 68.7
  p99: ......................................................................... 295.9
```
Not a single failed request!

## Testing

The test suite needed a minor change to work with the singleton engine object. I don't fully understand what the issue was here, but if the tests use the same `create_async_engine` params as the production app, the tests appear to try to reuse connections from the pool while other tests are still running. Overriding the engine creation in the test suite with a `create_async_engine(..., poolclass=NullPool)` arg solves the issue and allows the tests to work with the new setup.


garden-backend-service/src/api/dependencies/database.py
```diff
@@ -15,6 +21,52 @@
 SEARCH_SQL_INIT_LOCK_ID = 20250117  # YYYYMMDD when implemented
 
 
+class DatabaseEngine:
```
WillEngler, line 24: This looks like the right pattern for us to move to. Thanks for identifying the bottleneck, and the solutions looks clean too

WillEngler (approved):
LGTM!

PR #236, opened 2025-02-06
add example_usage field to modal functions

OwenPriceSkelly:
Follow up to #232 

## Overview

This gives modal functions a new example_usage field which is pre-populated based on test_functions extracted during the file validation step. 

## Testing
No new tests besides testing manually with a local instance of the frontend


garden-backend-service/src/api/schemas/modal/modal_function.py
```diff
@@ -12,6 +12,7 @@ class ModalFunctionMetadata(CommonFunctionMetadata):
     doi: str | None = None
     # modal supports conda requirements but entrypoints don't
     conda_requirements: list[str] = Field(default_factory=list)
+    example_usage: str = ""
```
WillEngler, line 15: I wasn't thinking originally about retaining both the original raw text we parsed out and the "cleaned up" text the user confirms. But this makes a lot of sense, nice

WillEngler (approved):
LGTM!

PR #239, opened 2025-02-11
fix: file validation image variable references

### LINKED ISSUE #238: fix: collect image info from references to other images
For example: 

```
BASE_IMAGE = modal.Image.debian_slim(python_version="3.11").pip_install(
    "my-packages"
)
FOO_IMAGE = BASE_IMAGE.run_function(download_foo_weights)

BAR_IMAGE = BASE_IMAGE.run_function(download_bar_weights)
```

this should be a relatively small change but we don't currently handle this case. at the very least it should fail gracefully instead of erroring

OwenPriceSkelly:
fixes #238 

## Overview
Quick QOL fix for an error I ran into making image variants for the full MACE garden. The edge case this fixes is probably more likely in a post-`@modal.build`-deprecation world than it was previously. 

## Discussion 

This was a pretty small change to the file validation module -- in addition to handling the extra base case, failing to parse a base image for a function now gives it the default image instead of erroring. I think this is reasonable since the image metadata is something we want to collect but which we aren't displaying to the user for review during the publication process.  

## Testing

No tests (yet) besides manual testing by publishing a garden locally

### DIFF
--- a/garden-backend-service/src/modal/user_file_parsing.py
+++ b/garden-backend-service/src/modal/user_file_parsing.py
@@ -48,7 +48,7 @@ class ModalImageInfo:
 def _image_default() -> ModalImageInfo:
     # helper for use with default_factory below
     call_node = ast.parse("modal.Image.debian_slim()").body[0].value
-    return try_parse_image_info(call_node)
+    return try_parse_image_info(call_node, parsed_images={})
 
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
LGTM!

PR #246, opened 2025-02-28
extract arxiv metadata from link 

### LINKED ISSUE #243: support extracting paper metadata from arXiv link

OwenPriceSkelly:
closes #243 
## Overview

Adds a route for extracting some metadata about a related paper from an arxiv.org link. The link can either point to the pdf on arxiv or the paper's regular page. 

## Discussion

I made a new pydantic base class for related metadata schemas so that they will include rather than ignore any fields they weren't explicitly expecting. More relaxed schemas make sense for this kind of metadata, since these are fields we persist as opaque json and don't _really_ care about the exact fields the user/external api has provided for us 

## Testing

manually with a few links I grabbed at random off of arxiv. I also added a couple tiny sanity-tests to make sure the arXiv ID is extracted correctly for both pdf and regular links

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,3 +178,4 @@ response.json
 /garden-backend-service/repopack-output.txt
 /garden-backend-service/.basedpyright
 /garden-backend-service/.envrc
+.envrc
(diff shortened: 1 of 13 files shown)


OwenPriceSkelly:
decided this made more sense on the frontend - kept schema tweaks in #247

PR #247, opened 2025-03-03
relax related work schemas

OwenPriceSkelly:
micro-PR with just the schema changes we want to keep from #246


WillEngler (approved):
LGTM!

PR #251, opened 2025-03-10
PATCH /gardens can update modal_function_ids 

OwenPriceSkelly:
## Overview

Turns out we didn't actually support updating the `modal_function_ids` attribute on existing gardens, since it didn't exist in the schema for PATCH requests (but does for creating new ones). oops!

## Testing

If I'm being completely honest I didn't test it at all yet due to the fact that this is just copy-pasting a block of existing entrypoints code. I'll eat a hat if this breaks anything 🤠


WillEngler (approved):
I wanna pick the hat

PR #253, opened 2025-03-13
fix f-string ast bug

OwenPriceSkelly:
## Overview

Fixes the bug riya discovered where an f-string in the `from_registry` method would cause validation to fail. 

Trying to reliably infer the base image when we hit an f-string node would be overly complex (and plausibly not possible to handle pathological cases). So instead of failing when we can't infer the exact base image, we can leave it at 'n/a'.

## Testing

Added a new unit test with the snippet of riya's code that surfaced the error in the first place


WillEngler (approved):
Good f"ix"

hholb:
Good stuff! I like the tests!

PR #254, opened 2025-03-17
add plural GET /modal-functions route

OwenPriceSkelly:
Another quick 🤠 PR to add a GET /modal-functions route with query params (instead of just GET /modal-functions/{id})

## Testing

Included some new unit tests and light manual testing seems sane. Not surprising since this was basically copy-pasted from the equivalent entrypoints route


hholb (approved):
LGTM! Love the tests!

PR #256, opened 2025-03-20
Make it so that a garden can be saved more than once

### LINKED ISSUE #267: Bookmarking is busted

WillEngler:
Fixes https://github.com/Garden-AI/garden-frontend/issues/267

## Overview

There were 409 (conflict) errors happening when you tried to save/bookmark a garden. It appears that the root cause here was a weird translation from ORM to SQL, leading in a primary key constraint we didn't want. The user<->garden relation table had a primary key constraint on garden. (So the first user to claim a garden is the only one able to save it!) What we wanted was a joint constraint on (user, garden) pairs. This PR updates the constraint to reflect that.

## Discussion

There is a warning when running the migrations from scratch to the effect of 

> garden-backend-service-api-1      | /app/migrations/versions/f293268b31c5_add_fields_to_users_table_createusers.py:24: SAWarning: Table 'users_saved_gardens' specifies columns 'user_id' as primary_key=True, not matching locally specified columns 'garden_id'; setting the current primary key columns to 'garden_id'. This warning may become an exception in a future release

I'm still mystified as to how the migration got created this way when it went fine for our other relation tables.

## Testing

I made sure the migration runs cleanly, and I plan on doing a final end to end test in staging.

### DIFF
--- a/garden-backend-service/src/models/_associations.py
+++ b/garden-backend-service/src/models/_associations.py
@@ -1,4 +1,4 @@
-from sqlalchemy import Column, ForeignKey, Integer, Table
+from sqlalchemy import Column, ForeignKey, Integer, PrimaryKeyConstraint, Table
 
 from src.models.base import Base
 
@@ -19,8 +19,9 @@ gardens_modal_functions = Table(
 users_saved_gardens = Table(
     "users_saved_gardens",
     Base.metadata,
-    Column("user_id", Integer, ForeignKey("users.id"), primary_key=True),
-    Column("garden_id", Integer, ForeignKey("gardens.id"), primary_key=True),
+    Column("user_id", Integer, ForeignKey("users.id")),
+    Column("garden_id", Integer, ForeignKey("gardens.id")),
+    PrimaryKeyConstraint("user_id", "garden_id"),
 )
 
 entrypoints_mdf_datasets = Table(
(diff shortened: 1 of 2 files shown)


garden-backend-service/src/models/_associations.py
```diff
@@ -19,8 +19,9 @@
 users_saved_gardens = Table(
     "users_saved_gardens",
     Base.metadata,
-    Column("user_id", Integer, ForeignKey("users.id"), primary_key=True),
-    Column("garden_id", Integer, ForeignKey("gardens.id"), primary_key=True),
+    Column("user_id", Integer, ForeignKey("users.id")),
+    Column("garden_id", Integer, ForeignKey("gardens.id")),
+    PrimaryKeyConstraint("user_id", "garden_id"),
```
WillEngler, line 24: Note: this did not "change" things from Alembic's perspective. I had to write the migration manually.

OwenPriceSkelly (approved):
LGTM ✅ 


Ive also noticed that sqlalchemy warning before when rebuilding my local docker compose images but didn't go down the rabbit hole. no insight besides that afaict it isn't really breaking anything

WillEngler:
> no insight besides that afaict it isn't really breaking anything

To be clear, I was trying to convey that that warning was a sign we could have noticed that indicated that the users_saved_gardens table was broken

PR #257, opened 2025-03-21
Eagerly load saved gardens to avoid async serialization error

WillEngler:
Fast follow to #256. I was testing in staging and noticed occasional 500's on saving operations. Logs looked like this:
```
[21/Mar/2025:16:22:19] INFO: 172.26.53.0:22590 - "PUT /users/6c9e223f-c215-4c26-9abb-262dbce0001c/saved/gardens/10.23677/kga9-nc58 HTTP/1.1" 200 OK
[21/Mar/2025:16:22:21] event="Saved garden" level=info doi=10.23677/m4ek-bd27 func_name=save_garden request_id=1615c3b4-73ec-4084-a50e-501fcad742c2 timestamp=2025-03-21T16:22:21.907435Z user_identity_id=6c9e223f-c215-4c26-9abb-262dbce0001c username=willengler@uchicago.edu
[...]

PR #260, opened 2025-03-21
Load even more eagerly

WillEngler:
Follow-up to #257. Still seeing async serialization errors on backend. This PR tries to even more explicitly eagerly load the user-garden relationship


WillEngler:
Gonna cowboy this to test/catch the async issues in staging

Issue #250, opened 2025-03-10
bump modal client version


OwenPriceSkelly:
we've been warned! our deployed apps will need to be redeployed along with the version bump. https://modal.com/docs/guide/modal-1-0-migration

PR #262, opened 2025-03-24
bump modal version to `0.73.116`

### LINKED ISSUE #259: refactor invocation of modal.Cls methods
https://modal.com/docs/guide/modal-1-0-migration#simplifying-cls-lookup-patterns 

we currently get away with treating methods as "functions with a `.` in their name", but going forward we will apparently need to instantiate the class itself once we look it up. according to the migration guide there may be "subtle differences" in behavior between the old and new way we should make sure we understand

OwenPriceSkelly:
fixes #259, first part of #250.

Leaving 250 open until 1.0 is officially here and we're compatible, but going by the migration guide I suspect we won't need to make any more major changes. 

## Overview

This PR handles the tricky parts of the modal 1.0 migration having to do with looking up and hydrating the `_Cls`/`_Function`/`_Invocation` modal objects. 

## Discussion

The "big picture" behind the necessary changes is that the modal library could load an object's metadata eagerly or lazily, and they're deprecating the eager methods. We had been able previously to use the eager `.lookup` method to fetch both regular `_Functions` and class methods, but we now need to fetch them lazily and explicitly force them to load the metadata they'll need for handling invocations. 

We also need to do a little extra work to access the `_Functions` underlying class methods, since the class metadata is now resolved independently from the method metadata (bc it's lazy). This means we have to create an instance of the user's class, call `getattr` with the function name, and then finally hydrate the `_Function` we get from the attribute so we can invoke it. 

## Testing

So far just manually by creating a toy garden with regular functions and class methods. I'll also want to sanity check a deployment after the lambda functions have been updated with the version bump

### DIFF
--- a/.github/workflows/deploy-lambdas.yml
+++ b/.github/workflows/deploy-lambdas.yml
@@ -43,7 +43,7 @@ jobs:
           docker run --rm -v $(pwd):/var/task public.ecr.aws/sam/build-python3.12 /bin/sh -c "\
              pip install --upgrade pip && \
              pip install watchfiles --platform manylinux2014_x86_64 --only-binary=:all: --target ./packages && \
-             pip install modal==0.66.37 --target ./packages"
+             pip install modal==0.73.121 --target ./packages"
           # Put the requirements into the top level of our deployable zip archive.
           cd packages
           zip -r ../app.zip *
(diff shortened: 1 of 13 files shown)


WillEngler (approved):
Looks good! Glad it wasn't too involved

PR #266, opened 2025-03-31
Allow volumes and memory snapshotting to be configured

WillEngler:
I'd like to use volumes to store the txGemma weights and memory snapshotting to speed up cold starts. It'll be more of a feature to make volumes useful to not just the dev team.


hholb (approved):
Nice!! Excited to mess with the snapshotting!

Issue #255, opened 2025-03-17
Redeploy Modal apps with updated client


WillEngler:
A follow-up to #250. Once we've bumped the client, we need to re-deploy our existing apps in place. This is probably also an ok time to purge test gardens that are lying around taking up our container quota without doing anything.

PR #269, opened 2025-04-01
app redeploy route

OwenPriceSkelly:
part of #255, implements the route that we can use to actually do the migration 

## Overview

Adds a new POST modal-apps/redeploy/{id} route that is only available to super users. This route reuses the deployment helpers from the regular app create flow, but doesn't change anything in the database -- the only actual changes are on modal's end since this will be using the new version of the client. 

## Testing

Some unhappy path unit tests, and on dev as soon as this is merged 🤠


WillEngler (approved):
Let's do it!

PR #271, opened 2025-04-02
Support filtering garden search results by owner/gardener

hholb:
## Overview

This PR adds support for filtering garden search results by 'Gardeners'.

## Discussion

Essentially just handles adding a where clause to the search filters query if the gardener/owner filter is in the request, and adds a `gardeners` field to the facets schema.

## Testing

Updated existing unit tests.

Tested manually on local FE.


WillEngler (approved):
Looks good!

PR #272, opened 2025-04-02
DELETE /modal-apps requests a delete on modal, too

### LINKED ISSUE #270: make DELETE /modal-apps actually deactivate the app on modal

OwenPriceSkelly:
closes #270 
## Overview

This plagiarizes the modal CLI command to stop an app and makes it so that `DELETE /modal-apps` attempts to stop the app on modal. This doesn't include any more sophisticated error handling / retry logic. We'll need to think harder about that anyways in the future when we actually run into our 200 app limit and implement a workaround 

## Testing

Updated the delete route's unit test. Still need to do some manual sanity-checking


WillEngler (approved):
Cool!

PR #273, opened 2025-04-02
fix modal request logic

OwenPriceSkelly:
oops! #272 didn't actually work when tested on dev, throwing up some familiar modal x asyncio errors. this seems to fix that by using the same `retry_transient_errors` helper to actually make the request that we've needed to use elsewhere.


hholb (approved):
Looks good!

PR #280, opened 2025-04-10
fix app id lookup

OwenPriceSkelly:
Gosh, I sure hope this works 

## Overview

This should be a safe workaround for the inconsistent errors we were seeing when deleting modal apps, all of which seemed to stem from failing to actually fetch modal's internal app id. 

This PR has two enhancements: 
  - we now persist the `modal_app_id` string whenever we (re-)deploy an app, so all new apps will be able to get the id directly from our database instead of a modal API call 
  - when we don't have the app id already present, we look it up by listing all of the apps in the environment and grabbing the id field from there. This was the only one of the ~3 plausible methods of looking up an app that seemed to actually work. 

## Testing

So far, just manually by creating a new app, observing that the ID was indeed persisted at deployment time, and then deleting the app. I'll need to cowboy this to dev in order to *really* test that we can now delete the previously un-deletable apps, but I was able to check that the `AppListResponse` had a populated id field for every app on our dev workspace so I think we're safe.


hholb (approved):
LGTM!

PR #281, opened 2025-04-10
Add route to find gardens using functions by function id

hholb:
This PR supports https://github.com/Garden-AI/garden-frontend/pull/327.

## Overview

This PR adds a route to find gardens using functions given a list of function ids.

## Testing

Manually

### DIFF
--- a/garden-backend-service/src/api/routes/gardens.py
+++ b/garden-backend-service/src/api/routes/gardens.py
@@ -81,19 +81,19 @@ async def search_gardens(
     return result.all()
 
 
-@router.post(
+@router.get(
     "/gardens-using-functions",
     status_code=status.HTTP_200_OK,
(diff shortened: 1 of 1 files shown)


garden-backend-service/src/api/routes/gardens.py
```diff
@@ -81,6 +81,34 @@ async def search_gardens(
     return result.all()
 
 
+@router.post(
+    "/gardens-using-functions",
```
WillEngler, line 85: I think this is a little wonky from a REST perspective. This is a retrieval, so it seems like it should be a GET. And /gardens-using-functions aren't distinct resources from /gardens. Can this functionality be added to GET /gardens with a query param?

WillEngler (approved):
We should make this a variant of GET /gardens because it's the same entity. But gonna merge this now and tweak later

PR #283, opened 2025-04-14
split invocations table

### LINKED ISSUE #282: split `modal_invocations` table into results vs usage
Our `modal_invocations` table currently has two distinct responsibilities: 
  - estimating per-user monthly usage (and any future "bookkeeping" analytics we may want) 
  - querying for the outputs/error/status of an asynchronous invocation 

```python
class ModalInvocation(Base):
    __tablename__ = "modal_invocations"
    id: Mapped[int] = mapped_column(primary_key=True)
    user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))   # bookkeeping
    function_id: Mapped[int] = mapped_column(ForeignKey("modal_functions.id")) # results 
    function_call_id: Mapped[str]  # results
    date_invoked: Mapped[datetime.datetime] = mapped_column( # bookkeeping 
        DateTime(), server_default=func.now()
    )  
    date_resolved: Mapped[datetime.datetime] = mapped_column(nullable=True) # bookkeeping 
    estimated_usage: Mapped[float] = mapped_column(default=0.0) # bookkeeping 
    status: Mapped[AsyncModalJobStatus] = mapped_column( # results 
        default=AsyncModalJobStatus.PENDING
    )  
    error: Mapped[str] = mapped_column(nullable=True) # results 
    output: Mapped[bytes] = mapped_column(nullable=True) # results 
```

We want to be able to delete modal functions (and their outputs) without losing all of their invocation history. 

We should separate `modal_invocations` into two tables - we can keep `modal_invocation_results` for _results_ of invocations, and introduce a new `modal_invocation_logs` table for bookkeeping, with the following relationships/behavior: 

  - `modal_functions` to `modal_invocation_logs` should be a _nullable_ one-many relation, so deleting a function does not delete its invocation history 
  - `modal_functions` to `modal_invocation_results` should be _non-nullable_ one-many, so that deleting a function _does_ cascade delete its invocation results 
  - `modal_invocation_logs` to `modal_invocation_results` should be a _nullable_ one-one, so that deleting invocations does not delete records (deleting records should delete invocations, if we ever need to delete records) 
  - `users` to `modal_invocation_logs` should be a _non-nullable_ one-many, so that we can calculate monthly usage

OwenPriceSkelly:
closes #282

## Overview

This PR handles separating our existing invocations table into two tables for results and for usage logging (More context in #282)

The important parts worth a closer review are the new DB models' relationships and the alembic migration script, which should migrate the existing invocation data to the new tables. The changes to the routes/business logic itself were fairly minimal.

## Testing

I've tested it manually locally and updated the unit tests (which did surface an actual mistake) but I haven't tested the migration logic itself against a non-empty DB, so I won't be convinced until I see the deployment succeed on dev.


WillEngler (approved):
LGTM!

PR #291, opened 2025-04-15
Consolidate searching by function id into existing /gardens endpoint

WillEngler:
Following up on #281. /gardens-using-functions isn't a distinct RESTful resource, so we need to consolidate it into /gardens


OwenPriceSkelly (approved):
LGTM

Issue #277, opened 2025-04-08
Backend support for updating a Modal app


WillEngler:
Probably something like `PUT /modal-apps/async/{id}`?

I don't know the exact route(s) we'll need yet, but we will need the ability to atomically replace a Modal app in place with a new version.

PR #295, opened 2025-04-17
Owen/update modal apps inplace

### LINKED ISSUE #292: users can update a modal app but not remove functions
First pass at support for updating a modal app in place can require that functions in the updated app are a superset of functions already existing in the app

OwenPriceSkelly:
fixes #292 (first part of #277)

## Overview

This PR allows users to update metadata of modal apps they own in-place with `PATCH /modal-apps/async/{id}`. If the `file_contents` field is included in the request body, we reproduce the "full publishing flow" as if we were hitting /modal-file-metadata for validation before re-deploying the app. 

## Discussion

We don't deploy the app if: 
  - the app's name has changed in the source code (bc we deploy the app under a fixed name this isn't technically an obstacle, but is a signal that the user probably picked the wrong file for the update) 
  - The functions defined in the new source code are not a superset of the functions in the old source code. The functions can be redefined as long as they have the same names, in which case their DB entry will be updated in-place. 
    - For now, this is true whether or not they are actually used by any gardens. Follow up PR will include the deletion logic to relax the constraint so that users can omit unused functions. 
## Testing

Manually + some new unit tests

### DIFF
--- a/garden-backend-service/src/api/routes/modal/modal_apps.py
+++ b/garden-backend-service/src/api/routes/modal/modal_apps.py
@@ -3,7 +3,7 @@ from uuid import uuid4
 
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, status
 from modal_proto import api_pb2
-from sqlalchemy import and_, select
+from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 from structlog import get_logger
 
@@ -34,7 +34,6 @@ from src.modal import parse_modal_file
 from src.modal.status import AsyncModalJobStatus
 from src.modal.utils import monitor_modal_deployment
 from src.models import ModalApp, ModalFunction, User
-from src.models._associations import gardens_modal_functions
 
 from .modal_file_metadata import parse_modal_file_metadata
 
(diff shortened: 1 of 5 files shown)


garden-backend-service/src/api/schemas/modal/modal_app.py
```diff
@@ -52,3 +52,10 @@ class ModalFileMetadataRequest(BaseSchema):
 class ModalFileMetadataResponse(ModalAppMetadata):
     # successful POST /modal-file-metadata response should have everything needed for the subsequent CreateRequest
     pass
+
+
+class ModalAppPatchRequest(BaseSchema):
+    base_image_name: str | None = None
+    requirements: list[str] | None = None
+    conda_requirements: list[str] | None = None
```
hholb, line 60: Are we still supporting conda? No issues if we are. I thought we had removed conda support, but I might be thinking of the docker publishing flow.

garden-backend-service/src/api/routes/modal/modal_apps.py
```diff
@@ -147,6 +152,118 @@ async def redeploy_modal_app(
...
+    # TODO: required_names only needs to be functions actually in use
+    required_names = set(existing_functions.keys())
+
+    # collect function names from the updated file_contents
+    parsed_metadata = await parse_modal_file_metadata(
+        ModalFileMetadataRequest(file_contents=patch_request.file_contents)
+    )
+    parsed_fn_names = {fn.function_name for fn in parsed_metadata.modal_functions}
+    # updated app must not remove any functions in use
+    if missing_names := required_names - parsed_fn_names:
```
hholb, line 206: This is slick, nice!

hholb (approved):
LGTM! Love the tests!!

garden-backend-service/src/api/schemas/modal/modal_app.py
OwenPriceSkelly: yeah we grab these from the `Image.micromamba_install` method if it appears during file validation, but you're right we don't really use/encourage them over regular pip requirements. might be worth just taking them out altogether

PR #297, opened 2025-04-21
app updates can omit unused functions to delete them

### LINKED ISSUE #277: Backend support for updating a Modal app

### LINKED ISSUE #293: users can update a modal app and omit unused functions
second pass at updating an app in place should allow users to omit functions from the app as long as they aren't referenced by any gardens

OwenPriceSkelly:
closes #293 (is the second half of and closes #277)

## Overview

This is the second pass at `PATCH /modal-apps`, which now fully handles cases where a user has omitted functions from a new app's definition. If the functions are safe to delete (not referenced by any gardens) they are deleted, otherwise a helpful error message is raised containing the functions that are still referenced by gardens, along with a list of the gardens (by title and doi) which are blocking the update. 

## Discussion

The diff is a little bigger than it strictly needed to be, but I think the route was getting awkwardly long so I cleaned it up a bit / factored out some helper functions. 
## Testing

manually sanity checked, plus included an update to an old test and a new unhappy-path test

### DIFF
--- a/garden-backend-service/src/api/routes/modal/modal_apps.py
+++ b/garden-backend-service/src/api/routes/modal/modal_apps.py
@@ -3,7 +3,7 @@ from uuid import uuid4
 
 from fastapi import APIRouter, BackgroundTasks, Depends, HTTPException, status
 from modal_proto import api_pb2
-from sqlalchemy import select
+from sqlalchemy import and_, select
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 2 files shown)


hholb (approved):
LGTM!

PR #299, opened 2025-04-24
Persist deployment output and suggested fixes during deployments

### LINKED ISSUE #276: Modal deployment failures should report more thorough error information back to the caller
When Modal deployments fail, I usually need to go check the logs to see what happened. To the extent it's safe/reasonable, we should be forwarding this information to the caller so that the caller can act on it (eg, to edit the Modal file)

hholb:
Resolves: #276 

## Overview

This PR upgrades the deployment handling to capture and persist the output from modal, as well as our generated suggested fixes, if an error occurs during deployment of a modal app.

## Discussion

Adds `suggest_fix` and `deployment_output` fields to the modal app table as well as the async modal app responses.

The biggest changes are to the `monitor_modal_deployment` background task, and the lambda function. None of the underlying logic changed, just added more robust error handling. The lambda now captures stdout from modal during deployment and returns it to the backend along with any errors.

## Testing

Manually since we don't have a great way to trigger the lambdas that generate these errors in the test suite.


garden-backend-service/src/modal/utils.py
```diff
@@ -93,18 +96,79 @@ async def monitor_modal_deployment(
...
+            exception_data = e["ModalException"]
+            suggested_fix = exception_data.get("suggested_fix")
+            deployment_output = exception_data.get("deployment_output")
+        else:
+            # Add suggested_fix based on error message
+            error_str = str(e).lower()
+            if "image build" in error_str or "failed with the exception" in error_str:
+                suggested_fix = "Try running the file locally with `modal run <filename>.py` to debug the issue."
+            elif "timeout" in error_str:
+                suggested_fix = "The deployment took too long. Try simplifying your container setup or breaking it into smaller parts."
```
WillEngler: I'm not sure this is actually the advice we want to give. Ideally they just wait a minute and try again

WillEngler (approved):
Overall this is good, but I think we should avoid telling users to scale down their container builds

garden-backend-service/src/modal/utils.py
hholb: Good catch!

Issue #284, opened 2025-04-15
Create auto-deletion task


WillEngler:
Set up hourly (?) background task that tags and removes unloved gardens and model deployments per the design doc

PR #302, opened 2025-04-30
Mark stale gardens and modal apps for deletion

hholb:
Part one of #284 

## Overview

This PR is the first stage of the auto deletion logic. Here, I setup the structure for the backgroud task, and add logic to mark gardens and modal apps with a timestamp for deletion.

**What this does:** 
- Updates database models and response schemas to support auto deletion
- creates a new auto deletion background task that starts during application startup and runs forever in the background
- the background task marks entities that are candidates for deletion with a timestamp

***What this doesn't do:***
- delete anything yet
- unmark object that are no longer deletion candiates

## Discussion

For the timestamp, I added a field called `marked_for_deletion` to both `ModalApps` and `Gardens`. I included this field in the response schemas as well, since I figure we will want to know what things are marked on the FE.

I created a new long-running background task, called `auto_deletion_background_task` that gets started during application startup, using the lifetime parameter for FastAPI. The task loops forever in the background, sleeping between iterations with the sleep duration configurable with a new setting `AUTO_DELETION_INTERVAL_SECONDS`.

On each iteration, the background task queries the database for draft gardens and unused modal apps, adds a timestamp to the `marked_for_deletion` field for any it finds. Then, is setup to delete entities that are past the age limit for marked objects, the age limit is configurable with a new `AUTO_DELETION_AGE_LIMIT_DAYS` setting. Nothing happens in the deletion functions yet, that will come in the next PR.

## Testing

New tests!!

An we finally have a way to interact with a db session directly in tests without having to go through the API routes!! Checkout the new `async_db_session` fixure which lets us create and use sessions within tests.

I added the [Faker](https://faker.readthedocs.io/en/master/index.html) library as a dev dependency to make is easier fill out fields of db models when we don't care what is in them.


garden-backend-service/tests/sandboxed_functions/test_modal_publishing_helpers.py
```diff
@@ -18,7 +18,7 @@ def hello_with_specs():
 
 def test_get_function_specs_raises_on_invalid_key():
     """Confirms that we will get errors when Modal changes their `_FunctionSpec` schema"""
-    invalid_specs = ["some", "specs", "that" "dont", "exist"]
+    invalid_specs = ["some", "specs", "thatdont", "exist"]
```
WillEngler, line 21: 🤔

garden-backend-service/tests/api/tasks/auto_deletion/test_utils.py
```diff
@@ -0,0 +1,185 @@
...
+        assert garden.marked_for_deletion is not None
+        marked_time = garden.marked_for_deletion
+
+        num_marked_second_time = await mark_entity_for_deletion(
+            Garden, async_db_session
+        )
+        assert num_marked_second_time == 0
+        await db.refresh(garden)
+        assert (
+            garden.marked_for_deletion == marked_time
```
WillEngler, line 184: Nice!!

garden-backend-service/src/main.py
```diff
@@ -55,6 +49,24 @@ async def startup_event():
     if settings.GARDEN_ENV in ["dev", "local"]:
         app.include_router(load_test.router)
 
+    # kick off long-running auto-deletion task
+    # dont await it, we want it to keep running while we move on
+    asyncio.create_task(auto_deletion_background_task(settings, session_maker))
+
+    # Everything before this yield happens on startup
+    yield
+    # Eveything after the yield happens on shutdown
```
WillEngler, line 58: I appreciate these comments, this would be obscure to me otherwise

WillEngler (approved):
This looks great, I'm excited for the sequel

PR #303, opened 2025-05-01
Delete objects marked for deletion

### LINKED ISSUE #284: Create auto-deletion task

hholb:
Auto-deletion part 2
Resolves: #284 

## Overview

Building off #302, this PR adds logic to unmark marked entities if they become non-deletion candidates before the time limit, as well as logic to actually delete entities that are past the time limit.

## Discussion

Nothing too crazy here, just adding to the structure setup in #302.

One extra point of complexity here was stopping modal apps on modal's end before deleting them from our database. I pulled the `stop_modal_app` helper out of the route file and into the modal utils so we can reuse it in the background task.

## Testing

New unit tests that confirm the background task unmarks marked objects correctly, and deletes objects when they have been marked for long enough while making sure we stop modal apps before deleting them from the db.

### DIFF
--- a/garden-backend-service/src/api/tasks/auto_deletion/utils.py
+++ b/garden-backend-service/src/api/tasks/auto_deletion/utils.py
@@ -74,3 +74,22 @@ async def unmark_marked_gardens(session_maker) -> int:
                 count += 1
         await db.commit()
     return count
+
+
+async def unmark_marked_modal_apps(session_maker) -> int:
+    # Subquery for all used modal function IDs
+    used_function_ids = select(gardens_modal_functions.c.modal_function_id)
+    # ModalApps where marked_for_deletion is not None and any of their functions are in use
+    stmt = select(ModalApp).where(
+        ModalApp.marked_for_deletion.isnot(None),
+        ModalApp.modal_functions.any(ModalFunction.id.in_(used_function_ids)),
+    )
+    count = 0
+    async with session_maker() as db:
+        results = await db.scalars(stmt)
+        marked_apps = results.all()
+        for app in marked_apps:
+            app.marked_for_deletion = None
+            count += 1
+        await db.commit()
+    return count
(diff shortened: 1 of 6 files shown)


hholb:
Not sure why CI didn't run on my last push, but I ran the tests job manually and it passed!

garden-backend-service/src/api/tasks/auto_deletion/utils.py
```diff
@@ -49,13 +91,51 @@ async def mark_gardens_for_deletion(session_maker):
 
 
 async def mark_modal_apps_for_deletion(session_maker) -> int:
-    # First get all modal function IDs that are used in gardens
-    used_function_ids = select(gardens_modal_functions.c.modal_function_id)
+    # First get all modal function IDs that are used in published gardens
+    used_function_ids = (
+        select(gardens_modal_functions.c.modal_function_id)
+        .join(Garden, Garden.id == gardens_modal_functions.c.garden_id)
+        .where(Garden.doi_is_draft.isnot(True))
```
WillEngler: I think this implies that we won't delete apps used in archived gardens. I think we also want to delete apps where doi_is_draft is False, but is_archived is True. (ie, the garden is archived)

garden-backend-service/src/api/tasks/auto_deletion/utils.py
```diff
@@ -49,13 +91,51 @@ async def mark_gardens_for_deletion(session_maker):
...
+            if not g.doi_is_draft:
+                g.marked_for_deletion = None
+                count += 1
+        await db.commit()
+    return count
+
+
+async def unmark_marked_modal_apps(session_maker) -> int:
+    # Subquery for all used modal function IDs
+    used_function_ids = select(gardens_modal_functions.c.modal_function_id)
```
WillEngler: And it seems like the unmark operation isn't symmetrical. (I might be missing something but it looks like the marking logic would mark an app for deletion if it's used in an unpublished garden, and then this unmarking logic would unmark it right after. I don't see this conditioning on the state of the garden)

WillEngler (commented):
The double boolean regime is confusing and I'd like to clean it up, but for now we should be only be treating gardens where `doi_is_draft is False` and `is_archived is False` as "published". I think the current logic lumps archived gardens into the published category

WillEngler (approved):
This looks great!

PR #307, opened 2025-05-05
do the same thing to the results table

WillEngler:
Part 2 of hotfix. This worked for the logs table, time to do it to its cousin


OwenPriceSkelly (approved):
LGTM

PR #312, opened 2025-05-13
fix unescaped '/' in url from doi string

OwenPriceSkelly:
This fixes the bug ben surfaced earlier where the doi.org redirect was sending users to the incorrect gardens.ai url. It also removes the old `/doi` routes now that the FE no longer uses them.


hholb (approved):
LGTM!

PR #313, opened 2025-05-14
create inital benchmarking models and routes

hholb:
First pass at BYOB features.

## Overview

This PR sets up the inital data model for the BYOB system, as well as defining the API routes and schemas to:
- Get data about available benchmarks
- Request a benchmark run against a function
- Get the run metrics for a given benchmark

**What this doesn't do**:
- Provide a way to create a new benchmark through the API
- Do any real validation that the function being benchmarked is compatible with the benchmark
- Probably some other stuff that hasn't come to mind yet

## Discussion

This adds three new entities to our database:
- `Benchmark` -- represents the concept of a Benchmark
- `BenchmarkTask` -- a specific task that is part of a benchmark, this is what maps to a function we call to run the benchmark
- `BenchmarkRun` -- a specific run of a benchmark task against a function

The idea is that a `Benchmark` is an entity that can hold some metadata about the nature of the benchmark. A `BenchmarkTask` is a specific task that is part of a benchmark, for example with Matbench Discovery there is a an IS2RE task and a S2EF (I think). These tasks are what we map to a function that we call to run the benchmark.

So the high-level idea is that a `BenchmarkTask` is a garden function that takes another garden function as an argument, runs whatever code that task has using the passed in function, then returns a dictionary of metrics. We store the invocation outputs just like any other invocation result, the `BenchmarkRun` object tracks which invocation record corresponds with the benchmark, task, and function being benchmarked.

This PR adds 3 new routes at `/benchmarks`.
- GET `/benchmarks` -- returns a list of available benchmarks and their tasks
- POST `/benchmarks/{benchmark_id}/{task_id}` -- requests a new run of the given benchmark task, the id of the function being benchmarked is in the request body
- GET `/benchmarks/{benchmark_id}/{task_id}` -- returns a list of benchmark results, one result for each function that has been benchmarked using the requested task.

## Testing

Unit test covering the basic flow of creating and defining a benchmark, then running the benchmark through the API.

Manually


OwenPriceSkelly (approved):
LGTM!

Issue #241, opened 2025-02-12
enhancement: catch module level imports on first validation pass


hholb:
I have bitten by this a couple of times now, even knowing to watch out for it!

We can't allow module level imports in modal apps since we don't have a good way of sending local files along with the modal app. Any module level imports other than modal and std lib modules  cause the deployment to fail when trying to pull the `modal.App` out of the the file.

It is a pretty easy fix, just move any imports into function scope and re-upload the file. 

The biggest pain point here is that we don't see this problem until after filling out the form with function titles, descriptions, and examples, then getting a failing deployment. Uploading a new version of the modal app clobbers all of the previously entered text.

I propose we look for module level imports during the initial file validation pass and warn users of this before they fill out the rest of the form.

hholb:
Closed by #318

PR #318, opened 2025-05-20
improve handling of modal app validation and deployment errors

### LINKED ISSUE #310: Obscure error message on attempted Modal app deployment
I tried uploading a Modal file that had an import in the global scope like 

```
import modal
from parse_cifs import read_experimental_cif
```

Obviously it's gonna fail on that second import. But the error message gave no indication of whether this was a problem I needed to fix in the Modal file or a problem on Garden's side.

Image

hholb:
Resolves: #310 #241 

## Overview

This PR fixes a couple of things, see the #310 comments for details about the CORS issue this fixes. The other fix is an enhancement we have been wanting for a while: catching module level imports during the initial validation pass instead of waiting for modal to throw an error during deployment. This means it shouldn't be possible to get to a state where the error Will found in #310 shows up, but I added some improved handling of `ModalException`s just in case.

## Discussion

The CORS issue is fixed by making sure we include the CORS headers when returning a `JSONResponse` after parsing `ModalException`s.

I added a check during the validation flow to raise an error if we find any module-level imports that are not `import modal`.

## Testing

New unit tests!
Manually!

### DIFF
--- a/garden-backend-service/src/exceptions/modal.py
+++ b/garden-backend-service/src/exceptions/modal.py
@@ -29,4 +29,9 @@ def handle_modal_exception(request: Request, error: ModalException):
     return JSONResponse(
         status_code=error.status_code,
         content=content,
+        headers={
+            "Access-Control-Allow-Origin": "*",
+            "Access-Control-Allow-Methods": "*",
+            "Access-Control-Allow-Headers": "*",
+        },
     )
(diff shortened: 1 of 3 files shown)


OwenPriceSkelly (approved):
LGTM! love the tests as always

PR #319, opened 2025-05-21
allow module-level imports from std library

hholb:
Related: #310 #241 #318 

## Overview

In #318 we added a check during the initial validation of modal apps to alert users if there are module-level imports. After merging into dev, I discovered that this is blocking imports from the python std lib which is unnecessary. This PR allows imports from the std lib to pass validation.

## Testing

A couple new unit tests!

### DIFF
--- a/garden-backend-service/src/modal/user_file_parsing.py
+++ b/garden-backend-service/src/modal/user_file_parsing.py
@@ -1,5 +1,6 @@
 import ast
 import copy
+import sys
 from dataclasses import dataclass, field
 
 import modal
@@ -81,14 +82,15 @@ class ModalFileParseResults:
 
 
 def _validate_module_level_imports(tree: ast.Module) -> None:
-    from .utils import is_stdlib_module  # moved import here to avoid circular import
-
     """Check that there are no module-level imports other than 'import modal' or stdlib imports."""
     for node in tree.body:
         match node:
             case ast.Import(names=names):
                 for alias in names:
-                    if alias.name != "modal" and not is_stdlib_module(alias.name):
+                    if (
+                        alias.name != "modal"
+                        and alias.name not in sys.stdlib_module_names
+                    ):
                         raise ModalException(
                             detail=f"Module-level import '{alias.name}' is not allowed.",
                             suggested_fix="Place all imports other than 'import modal' or stdlib imports inside function or class scopes.",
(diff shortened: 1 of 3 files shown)


garden-backend-service/src/modal/utils.py
```diff
@@ -229,3 +231,17 @@ async def stop_modal_app(
         source=api_pb2.APP_STOP_SOURCE_PYTHON_CLIENT,
     )
     await retry_transient_errors(modal_client.stub.AppStop, stop_request)
+
+
+def is_stdlib_module(module_name: str) -> bool:
```
OwenPriceSkelly: I think `return module_name in sys.stdlib_module_names` should be equivalent to this check, and I'm a little wary of the importlib module due to my own lack of understanding of its nuances -- e.g. I couldn't tell you off the top of my head when we would expect `spec is None or spec.origin is None` to be true, nor when `spec.origin in ("built-in", "frozen")` would be the case

garden-backend-service/src/modal/user_file_parsing.py
```diff
@@ -81,21 +81,26 @@ class ModalFileParseResults:
 
 
 def _validate_module_level_imports(tree: ast.Module) -> None:
-    """Check that there are no module-level imports other than 'import modal'."""
+    from .utils import is_stdlib_module  # moved import here to avoid circular import
```
OwenPriceSkelly: what's the circular import? glancing at the imports here and in `.utils` I'm not seeing one, but if there really is then that's a smell I think we should deal with

OwenPriceSkelly (changes requested):
I've got a couple of worries, but I think the thrust of this is reasonable. Another option here I'm wondering about (since there are _some_ modules outside the stdlib that would still be importable by default) is to have non-standard imports just return a warning to the FE instead of blocking the upload entirely

garden-backend-service/src/modal/utils.py
hholb: Ooh! `sys.stdlib_module_names` is handy! Thanks, that makes this a lot easier to reason about.

garden-backend-service/src/modal/user_file_parsing.py
hholb: Got rid of the helper function, no longer an issue!

OwenPriceSkelly (approved):
LGTM

PR #321, opened 2025-05-29
fix: modal invocations not resolving

### LINKED ISSUE #320: bug: failed modal invocations sometimes don't get caught by the monitoring background task
While testing the benchmark system I have encountered this issue a few times:

I invoke a function, the function fails for some reason on modal's end, but the `monitor_modal_invocation` background task doesn't catch the failure and the invocation stays marked as `pending` in our database.

This manifests on the new benchmarks page as an unwavering notification about running benchmarks tasks (when the benchmark task function is the one that didn't get resolved correctly):

Image

hholb:
Fixes: #320

## Overview

This PR fixes an issue with the `monitor_modal_invocation` background task where sometimes a failed function invocation wouldn't get caught and was marked 'pending' forever.

I added a query param to the get route for benchmark results called `include_failed` to get better visibility into failed benchmark runs on the FE. My intention is to keep failed runs out of view of users, but since garden devs will likely need to do some debugging when running benchmarks, having this will be nice. I added it in this PR because I needed it to manually test and make sure we were catching failed benchmark invocations correctly, which is where we first saw this issue.

## Discussion

Pretty hefty refactor of the background task. The major highlights:

- restructured the main task `monitor_modal_invocation` to be more readable and have better handling of database sessions
- pulled the logic that parses raw invocation output from modal into a  (mostly) pure helper function
- Make the task handle all errors that might happen during the task to keep errors from propagating up and killing the server

While refactoring the raw invocation processing, I realized I had made a false assumption about hanlding outputs and errors from modal in my first pass. Specifically, I was assuming if we had any outputs at all from `pop_function_call_outputs` that the invocation succeeded and marked it as 'done'. 

Turns out, when an invocation has a runtime error we still get outputs (containing error info), but also get an exception field we can use to report what when wrong. I beefed up the handling to make sure we correctly determine whether the invocation actually failed, but still has outputs, and capture the exception information in the database.

I was also assuming we would always get a 'result' field in the output response, and that is not always the case. I think this was the root cause of forever 'pending' issue.

## Testing

The load test `test_concurrent_modal_invocations` in `tests/load_test.py` covers most of the background task logic.

I tested manually with some functions that cause errors and couldn't trigger a forever  'pending' state. Once this is in dev, I want to run some tests of Owen's matbench benchmark since that is where we first noticed #320

### DIFF
--- a/garden-backend-service/src/modal/utils.py
+++ b/garden-backend-service/src/modal/utils.py
@@ -98,7 +98,9 @@ async def monitor_modal_invocation(
         result = await ModalInvocationResult.get(session, id=db_result_id)
     if result is None:
         # We don't have this result in the db, bail
-        log.info(f"modal invocation result with id {db_result.id} not found in database")
+        log.info(
+            f"modal invocation result with id {db_result_id} not found in database"
+        )
         return
 
     try:
--- a/garden-backend-service/src/modal/utils.py
+++ b/garden-backend-service/src/modal/utils.py
@@ -50,9 +50,9 @@ async def _process_modal_invocation(
     invocation: modal._functions._Invocation,
     timeout_seconds: float,
 ) -> ModalInvocationResult:
-    """Parse ouptut and errors from raw modal invocations.
+    """Parse output and errors from raw modal invocations.
 
-    Tell modal to cancel the invocation if it has been longer than
+    Tell modal to cancel the invocation if it has been longer than timeout_seconds
     """
     # create result to hold the parsed output data
     result = ModalInvocationResult(status=AsyncModalJobStatus.PENDING)
@@ -98,7 +98,7 @@ async def monitor_modal_invocation(
         result = await ModalInvocationResult.get(session, id=db_result_id)
     if result is None:
         # We don't have this result in the db, bail
-        log.info("modal invocation result with id {db_result.id} not found in database")
+        log.info(f"modal invocation result with id {db_result.id} not found in database")
         return
 
     try:
--- a/garden-backend-service/src/modal/utils.py
+++ b/garden-backend-service/src/modal/utils.py
@@ -46,6 +46,35 @@ async def resolve_modal_invocation(
         raise ValueError(f"No Modal Function with id: {id} found!")
 
 
+async def _process_invocation_completion(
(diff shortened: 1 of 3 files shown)


garden-backend-service/src/modal/utils.py
```diff
@@ -46,48 +46,90 @@ async def resolve_modal_invocation(
...
-            except Exception as e:
-                # Otherwise, write the error to the DB
-                result.error = str(e)
-                status = AsyncModalJobStatus.ERROR
-                await resolve_modal_invocation(result, status, session)
-                await session.commit()
+        result = await ModalInvocationResult.get(session, id=db_result_id)
+    if result is None:
+        # We don't have this result in the db, bail
+        log.info("modal invocation result with id {db_result.id} not found in database")
```
OwenPriceSkelly: 🔍 👀 
```suggestion
        log.info(f"modal invocation result with id {db_result.id} not found in database")
```

garden-backend-service/src/modal/utils.py
```diff
@@ -46,48 +46,90 @@ async def resolve_modal_invocation(
         raise ValueError(f"No Modal Function with id: {id} found!")
 
 
+async def _process_modal_invocation(
+    invocation: modal._functions._Invocation,
+    timeout_seconds: float,
+) -> ModalInvocationResult:
+    """Parse ouptut and errors from raw modal invocations.
```
OwenPriceSkelly: 🔍 👀 
```suggestion
    """Parse output and errors from raw modal invocations.
```

garden-backend-service/src/modal/utils.py
```diff
@@ -46,48 +46,90 @@ async def resolve_modal_invocation(
...
+
+    Tell modal to cancel the invocation if it has been longer than
```
OwenPriceSkelly:
```suggestion
    Tell modal to cancel the invocation if it has been longer than timeout_seconds
```

OwenPriceSkelly (approved):
couple of typos but LGTM!

garden-backend-service/src/modal/utils.py
hholb: Good catch!

Issue #322, opened 2025-05-30
Schema Issue: Modal App level schema may be invalid if multiple images are used


loganalt9:
While trying to sync correct Modal images to apps, I ran into a potential schema issue. In the case that a Modal app contains multiple different images then a single, canonical `base_image` does not exist. 

This assumption breaks in cases like issue #309 where the system fell back to a default image with Python 3.12 for all apps, even when another version was specified.

PR #323, opened 2025-05-30
fix: modal image detection

### LINKED ISSUE #309: Image detection misfires
I uploaded a Modal file with an image like this

modal.Image.debian_slim(python_version="3.11")

and it reported an image like this

Image

loganalt9:
Fixes: #309 

## Overview

This PR is a targeted quick fix for Modal app image detection. Previously when adding a Modal file to garden with an image such as `modal.Image.debian_slim(python_version="3.11")` the correct version of Python was not caught and would default to 3.12.6. 

This was due to `parse_modal_file_metadata` using the app level schema to grab the base image, which was not being correctly defined and resorting to `_image_default`. This targeted fix checks if any images were defined, if so, it uses the first image's base_image. If no images were detected, `_image_default` is used which produces the Python 3.12.6 that was always seen earlier.

## Discussion

This is a quick fix specifically for issue #309 but does not address an underlying issue. If a Modal app has multiple images used across different functions then having a base image for the whole app could cause issues (#322).

## Testing

I was able to upload a garden with different versions of Python and see those get caught by the fix. I uploaded a Modal file with image `modal.Image.debian_slim(python_version="3.10")` and saw these results: 
image

### DIFF
--- a/garden-backend-service/src/api/routes/modal/modal_file_metadata.py
+++ b/garden-backend-service/src/api/routes/modal/modal_file_metadata.py
@@ -15,8 +15,8 @@ from src.exceptions.modal import ModalException
 from src.modal.user_file_parsing import (
     ModalFileParseResults,
     ModalLocalEntrypointInfo,
+    _image_default,
     parse_modal_file,
-    _image_default
 )
 from src.models import User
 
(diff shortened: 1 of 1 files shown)


WillEngler (approved):
LGTM! Thanks for the thorough context in the PR description

PR #316, opened 2025-05-16
fix: cascade garden deletion to users_saved_gardens

hholb:
Resolves: #314 

This PR fixes the auto deletion background task by cascading garden deletions to the `users_saved_gardens` table.

PR #327, opened 2025-06-02
fix: auto deletion modal client issue

### LINKED ISSUE #324: bug: autodeletion background task 'no modal client provided'
While looking through the lightsail logs I noticed we are getting a 'no modal client provided' error when trying to stop/delete stale modal app deployments. Here are the logs:

Image

This isn't causing any issues outside of the background task since it skips over any errors, but this means we aren't actually cleaning up anything when the task runs.

hholb:
Resolves: #324 

## Overview

This PR fixes a bug in the auto-deletion background task that was causing the 'no modal client provided' error we were seeing in the logs.

After fixing the error, I noticed another error along the lines of #316 where we weren't cascading deletes to the new benchmark tables, causing the database to reject the delete query sent by the background task. This PR fixes that issue as well.

## Discussion

Simple fix! When we added the modal client as an arg to the background task, we set a default of `None` and then didn't pass a client in at the call site, which lead to the 'no modal client' errors. Now we try to get a modal client on app startup, and the background task tries to create its own client if one is not provided.


garden-backend-service/src/api/tasks/auto_deletion/auto_deletion.py
```diff
@@ -31,6 +31,21 @@ async def auto_deletion_background_task(
 
     while True:
         logger.info("Auto-deletion task starting sweep...")
+
+        # If no modal client was provided at startup, try to create one now
+        current_modal_client = modal_client
```
OwenPriceSkelly: I'm not sure I follow what case we're handling here -- is there a reason we _wouldn't_ want to pass a modal client to the background task at startup? especially since we only ever call this once from the lifespan function, I think we could simplify and just make modal_client a required arg

garden-backend-service/src/main.py
```diff
@@ -49,9 +53,20 @@ async def lifespan(app: FastAPI):
     if settings.GARDEN_ENV in ["dev", "local"]:
         app.include_router(load_test.router)
 
+    # Try to create modal client for auto-deletion task
+    # If this fails, the task will retry on each sweep
```
OwenPriceSkelly: if we fail to create a modal client here, is there a reason to think it would succeed on a later sweep? or is the retry logic because the client gets "stale" after a certain point?

OwenPriceSkelly: I'm kinda thinking that if we can't instantiate a modal client, I think that we should fail so we can investigate why rather than suppress the exception

garden-backend-service/src/models/benchmark.py
```diff
@@ -31,7 +31,9 @@ class BenchmarkTask(Base):
         back_populates="tasks", lazy="selectin"
     )
 
-    function_id: Mapped[int] = mapped_column(ForeignKey(ModalFunction.id))
+    function_id: Mapped[int] = mapped_column(
```
OwenPriceSkelly, line 34: I'm not sure we want this cascade behavior for benchmark tasks/runs -- if an app containing a benchmarked function is stale and we clean it up, wouldn't this cascade mean that we lose the benchmark results? imo "exists on a benchmark" is a good reason for an attempted delete to fail

garden-backend-service/src/api/tasks/auto_deletion/auto_deletion.py
```diff
@@ -31,6 +31,21 @@ async def auto_deletion_background_task(
...
+        if current_modal_client is None:
+            try:
+                from src.api.dependencies.modal import get_modal_client
```
OwenPriceSkelly: is this a circular dependency? If so I think that points very strongly towards passing the client in as a required arg from the lifespan function so we don't need to import anything new here

OwenPriceSkelly (changes requested):
thanks for jumping on this! Couple things I'm not sure about (I may be missing something subtle), happy to sync up

garden-backend-service/src/models/benchmark.py
```diff
@@ -50,21 +50,21 @@ class BenchmarkRun(Base):
     )
 
     task_id: Mapped[int] = mapped_column(
```
OwenPriceSkelly: I believe we'll also need the annotation to read `Mapped[int | None]` and/or pass `nullable=True` kwarg to the mapped_column (like the regular `ModalInvocationLog/Result` models), otherwise iirc the cascade will try to set null but fail since the column isn't nullable

garden-backend-service/migrations/versions/02c0be279ee6_handle_function_deletion_in_related_.py
OwenPriceSkelly, line 1: I don't think this script actually picked up the changes to the models - there's no mention of the benchmark tables anymore

OwenPriceSkelly: I think this has happened to me before when my local alembic state got screwy, iirc deleting the migration scripts then recreating them after a fresh `docker compose up --build` fixed it

garden-backend-service/src/models/_associations.py
```diff
@@ -13,7 +13,11 @@
     "gardens_modal_functions",
     Base.metadata,
     Column("garden_id", ForeignKey("gardens.id"), primary_key=True),
-    Column("modal_function_id", ForeignKey("modal_functions.id"), primary_key=True),
+    Column(
+        "modal_function_id",
+        ForeignKey("modal_functions.id", ondelete="CASCADE"),
```
OwenPriceSkelly, line 18: I'm a little worried about this -- I think we still want the behavior of "if referenced by a garden, we can't delete the function", especially if we're doing the deletes in an automated fashion

OwenPriceSkelly: do we need to commit to this one way or the other in this PR? I feel like this is maybe something to revisit when we're coming back to the 'basket of deployables' problem

OwenPriceSkelly (changes requested):
the task logic looks much cleaner but I think there are still some gotchas with the models/tables. I'm also not sure about the cascade behavior on the gardens_modal_functions table -- happy to sync up if I'm just confused!

garden-backend-service/src/models/_associations.py
hholb: Our initial approach ([design doc](https://docs.google.com/document/d/1PSdpQ34zo4J8edltHPzzzFtxCnqNH5hs3mxBS7XAu-s/edit?tab=t.0#heading=h.leu1naawcdy2)) was to delete any garden in draft state, and any modal apps that aren't used by *published gardens*. We can have records in this table for draft and archived gardens, without the cascade we can't delete the modal apps without manually deleting records from this table first.

OwenPriceSkelly (approved):
LGTM! and the bike shed is *just* the right color

PR #329, opened 2025-06-05
Owen/benchmark args fn names

OwenPriceSkelly:
part of https://github.com/Garden-AI/uploadathon/issues/18 
## Overview

In order to use modal's batch invocation bells/whistles, we'll need to call the function to be benchmarked directly from modal (instead of via garden). 

Fortunately this is just a change to the arguments we're passing to the benchmark task, the rest of the changes to the logic will be in the modal app for the benchmark

## Discussion

We should hold off on merging this until I've got the updated modal app for the benchmark ready; it's nice that we can currently run the "small" version of the task against a function to see if the function is compatible and this would break that temporarily. 

## Testing

n/a


hholb (approved):
LGTM!

PR #335, opened 2025-06-10
Owen/fix base image parse

### LINKED ISSUE #334: image detection fix (the sequel)
@keqinghe2 saw the following behavior on staging, where base image found in the deploy step was different to the one found by the validation step. 

Image

This was a surprise outcome of #323, which corrected the image detection in the file validation route but not in the deploy route. We hadn't seen this behavior before now because both routes faithfully reproduced the same bug, i.e. it appeared to both as if the detected image was always the default python 3.12 (regardless of the actual image).

OwenPriceSkelly:
fixes #334 

## Overview

This just moves the logic from @loganalt9's recent PR from the route function to inside the `parse_modal_file` helper function, so that both the validation step and the final deployment step will find the same image (and the correct one now, too). 

## Discussion
There are some opportunities for us to clean up the file validation logic now that we know more about the files we can actually deploy with modal -- for instance, we could assume/enforce that there only is one app per file to simplify some of the parsing logic. That said, I decided not to be a good boyscout here because I don't think it'll slow us down next time we need to touch the file validation logic. 

## Testing

Manual smoke test by deploying an app locally

### DIFF
--- a/garden-backend-service/src/api/routes/modal/modal_file_metadata.py
+++ b/garden-backend-service/src/api/routes/modal/modal_file_metadata.py
@@ -15,7 +15,6 @@ from src.exceptions.modal import ModalException
 from src.modal.user_file_parsing import (
     ModalFileParseResults,
     ModalLocalEntrypointInfo,
-    _image_default,
     parse_modal_file,
 )
 from src.models import User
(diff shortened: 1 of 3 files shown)


WillEngler (approved):
LGTM!

PR #337, opened 2025-06-12
Remove /mdf/search route

### LINKED ISSUE #336: Remove unused /mdf/search route
AFAIK this is not being used, and it is getting in the way of Logan's MCP work. I think we can just axe it

WillEngler:
Closes #336. This helps clear the way for Logan deploying an MCP layer on top of the Garden API. One of the Pydantic models for Globus Search interfered with the MCP plugin library. AFAIK the /mdf/search route was not in active use, so it's safe to cut the whole route. We have more trimming to do still if we want to remove the rest of the feature (eg, database records)


hholb (approved):
LGTM! It's beautiful 😢

Issue #330, opened 2025-06-06
MCP hello world - access for metadata routes


WillEngler:
There is not a straightforward way to let LLMs call Garden models directly, but there are likely [off-the-shelf plugins](https://github.com/tadata-org/fastapi_mcp) we can use to expose Garden's metadata routes to LLMs. 

Success for this looks like the following: Claude can fetch Garden metadata via MCP. Someone using Claude Pro can call the Garden backend to search for a model, pull its metadata, and write some code that (largely) works to invoke the model. In this phase, the LLM does _not_ invoke a Garden model directly.

WillEngler:
I'd say we've achieved this in spirit

PR #338, opened 2025-06-12
Logan/mcp function metadata

loganalt9:
#330 

## Overview

This PR introduces an MCP server through the fastapi-mcp package. As of now this adds the ability for any MCP client, that is able to connect to a remote MCP server, to search through Garden to find Modal function metadata. In this version, only the `search_gardens` route has been mounted onto an MCP server.

## Discussion

If you want to try this out yourself with this code you will need a http tunneling service like ngrok. A public facing URL is necessary for remote MCP servers. Also, with Claude Pro, you can input the URL appended by the MCP route in settings and Claude will be able search for modal functions.

## Testing

Tested on Claude UI and Claude desktop, using search functionality to find functions by name, tags, authors, etc.

### DIFF
--- a/garden-backend-service/pyproject.toml
+++ b/garden-backend-service/pyproject.toml
@@ -32,7 +32,8 @@ dependencies = [
   "phonenumbers>=8.13.40",
   "structlog>=24.4.0",
   "sqlparse>=0.5.1",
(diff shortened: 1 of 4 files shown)


WillEngler (approved):
LGTM! Nice!

PR #339, opened 2025-06-16
modal 1.x update

### LINKED ISSUE #237: handle `gpu="L40S"` in modal usage calculations
In modal's 0.68.44 release they added support for a L40S gpu type, which is also the gpu their docs [recommend](https://modal.com/docs/guide/gpu#picking-a-gpu) as a starting point for running pre-trained models. 

This isn't _urgent_ as we're currently pinned to `modal<=0.67` for both the sdk and backend, but we'll want to make sure we cover this next time we upgrade.

### LINKED ISSUE #331: Upgrade to Modal 1.0
I got an email suggesting that some of our deployed apps may be deprecated soon. We need to ...

- Upgrade our Modal client in all parts of the system to at least 1.x
- Re-deploy any apps used in published Gardens with the new client

OwenPriceSkelly:
closes #331; also closes #237 along the way 

## Overview
Backend-side of updating to modal 1.x everywhere. This also bumps the lambda function's dependency, but the lambda function is pinned to 1.0.4 exactly (latest as of today). 

## Discussion

I'd already done most of the legwork for this the last time we bumped the modal dependency (i.e. we weren't using any of the now-deprecated internal methods) so this was also pleasantly very straightforward. 

After this we just need to redeploy the existing apps so modal stops emailing us about them. 

## Testing

Manual testing locally, but I'll want to double check that the lambda works on dev before we push all the way to prod/redeploy public gardens.

### DIFF
--- a/.github/workflows/deploy-lambdas.yml
+++ b/.github/workflows/deploy-lambdas.yml
@@ -43,7 +43,7 @@ jobs:
           docker run --rm -v $(pwd):/var/task public.ecr.aws/sam/build-python3.12 /bin/sh -c "\
              pip install --upgrade pip && \
              pip install watchfiles --platform manylinux2014_x86_64 --only-binary=:all: --target ./packages && \
-             pip install modal==0.73.116 --target ./packages"
+             pip install modal==1.0.4 --target ./packages"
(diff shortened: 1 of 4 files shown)


WillEngler (approved):
LGTM!

PR #340, opened 2025-06-23
MCP for function metadata pt 2

loganalt9:
#330 

## Overview

This PR fixes an incompatibility between fastapi-mcp plugin and Starlette's BaseHTTPMiddleware class that was internally called in our custom middleware. The added middleware `MCPBypassMiddleware` bypasses all other middleware if the request being made is to the mcp route, otherwise it sends every other request through the normal pipeline.

## Discussion

This is potentially a temporary fix, that might be reverted if we move away from the fastapi-mcp plugin completely. However as long as we are using that plugin it will conflict with our current middleware.

## Testing

Adding the `MCPBypassMiddleware` got rid of the assertion error that was shown in the logs when the previous mcp server was failing. Claude is also able to connect and search Garden for models based on our search route. I still need to test this remotely in staging, since last PR related to #330 also worked locally but threw an assertion error in prod.


OwenPriceSkelly (approved):
LGTM 🚢

PR #345, opened 2025-07-02
Logan/mcp function metadata

loganalt9:
#330 

## Overview

This PR moves away from the FastAPI-MCP plugin and to the MCP sdk's FastMCP class. Also, to work around an assertion error that arises in Starlette's BaseHTTPMiddleware this PR moves the logic from the existing middleware into pure ASGI middleware.

## Discussion

It's unclear if this switch will work before auth is implemented. The mcp server also now lives at /mcp/sse per the 2024-11-05 mcp spec.

## Testing

Locally Claude is able to make requests to the search garden tool and search via the same search parameters as the actual endpoint. New middleware also outputs expected/correct logs and handles errors in the same manner as the earlier middleware.


OwenPriceSkelly (approved):
LGTM! I mentioned already when we synced up earlier, but really nice work on the middleware -- very glad you found such a clean workaround. good stuff!

PR #353, opened 2025-07-11
mount mcp servers to test both transport types

OwenPriceSkelly:
This mounts two mcp server instances at different paths so that each one is testable -- since the [demo globus mcps](https://github.com/globus-labs/science-mcps/tree/6b447a23d3a1cf4cf6629fbffc17422e20f77b6f) were usable (from claude desktop using [mcp-remote](https://github.com/geelen/mcp-remote/blob/main/README.md)) when deployed on lightsail, I wanted to check and see if I could reproduce their success with our deployment (and also see if it was contingent on a specific transport type)

### DIFF
--- a/garden-backend-service/src/main.py
+++ b/garden-backend-service/src/main.py
@@ -1,12 +1,10 @@
 import asyncio
 import os
-import time
 from contextlib import asynccontextmanager
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


WillEngler (approved):
gun it

PR #358, opened 2025-07-30
Unique tags route

loganalt9:
Related to Garden-AI/garden#619

## Overview

This PR adds a new route at "gardens/tags" that aids the garden mcp search tool in supplying all the valid values for a `Garden.tags`. This is necessary to make sure mcp clients don't hallucinate invalid tag values when searching for gardens.

## Testing

Tested manually with Claude

### DIFF
--- a/garden-backend-service/src/api/routes/gardens.py
+++ b/garden-backend-service/src/api/routes/gardens.py
@@ -169,6 +169,13 @@ async def search(
     )
 
 
+@router.get("/tags", response_model=list[str])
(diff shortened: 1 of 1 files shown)


OwenPriceSkelly:
It occurs to me that we might not need this route if we hit the POST /gardens/search route instead, like the FE does when you use the search bar. I believe @hholb implemented it so that the search query syntax would be the same as [globus search](https://docs.globus.org/api/search/reference/post_query/)

PR #361, opened 2025-09-29
Create models and routes for HPC functions

### LINKED ISSUE #360: Design and take first stab at backend records for HPC functions
Our first pass at HPC functions in Edith is hardcoded. As we expand to Anvil (and hopefully more soon) I think it's time to formalize this a bit. This will be a featureset in transition as we continue to figure out HPC publication. But allowing for some manual data entry for now on the publication side, we can still work out things like:
- What DB objects should exist for HPC functions and deployments?
- How should these functions be invoked in the SDK based on data from the backend?
- How should we make non-hardcoded frontend pages from this data?

hholb:
Resolves: #360 

## Overview

This PR adds database models and basic CRUD routes for hpc-endpoints, hpc-functions, and hpc-deployments. I also updated the garden model and routes to allow associating hpc functions with gardens a la modal functions.

### The data model:
hpc-endpoints are just an id, name, and their globus-compute endpoint id.

hpc-deployments track the environment information for a set of functions. ~Deployments can either be uv (for upcoming groundhog support) and or conda envs (for our current mlip garden).~ I imagine this will change a bit as we work out the integration with groundhog and as MEPs add container support. Deployments can be associated with multiple endpoints, i.e. the mlip garden will be deployed on both edith and anvil.

hpc-functions share most of the same metadata as modal-functions including name, source text, and all of the associated materials (datasets, papers, etc.). Functions are associated with ~a single~ multiple deployments since deployments on different endpoints may have different setups for conda paths etc.

Note: This is a first-pass at the data model to get us up and running. It will likely need some tweaks as we get the SDK and frontend working with these entities.

### Creating new hpc functions:
1. Create an endpoint record `POST /hpc/endpoints` or use an existing one -- currently this is just behind a basic auth check, but we may want to restrict creating endpoints to super users.

2. Create a deployment record `POST /hpc/deployments` and associate it with endpoints where it has been created. Note: we don't handle creating the deployment environments, we only track their existence and the data we need to run functions.

3. Create a new function `POST /hpc/functions` and associate it with a deployment

4. Send a request to the `PATCH /gardens` route to add hpc functions to a garden

### Tracking invocations

Since hpc function invocation logic will be handled in the SDK, I created a basic invocation log table and a `POST` route the sdk can hit when an hpc function is invoked to create logs.

## Discussion

I refactored the function models to pull out common columns into mixin classes and moved both hpc and modal functions into a `/src/models/functions` directory -- this is what caused so many files to get touched, I needed to update imports anywhere that was importing modal functions.

## Future work
- I haven't implemented a route to create a deployment and a set of functions at the same time. This will need some more design and likely some file parsing that seemed out of the scope of this PR.
- Integrate HPC functions into the search features
- More robust update and delete logic

## Testing

Some basic unit tests of the new routes.
Manual testing using postman
Very basic data fetching in the SDK


OwenPriceSkelly (changes requested):
notes from sync: 

we decided we want to pare down to just the fields that the garden sdk will _need_ in order to function. since we don't know what we want to do with requirements etc we can cut them out ✂️ (but we did decide the mixins were cool enough to leave in) 

For deployments, this is: 
  - path to conda env 
  - function text and name 
    - groundhog at least (and possibly the frontend) would also want full original source text I think, but that can come later
  - known globus compute multiuser endpoints 
    - q: should we just derive this from invocation logs? maybe save a default endpoint? 
  - default user endpoint config (json)

For invocation logs, this is everything we have currently plus an _actual_ user endpoint config (json)

OwenPriceSkelly (approved):
LGTM! (with one tiny nit)

I think we could have the user_endpoint_config be non-nullable on both deployments and invocations, since we'll always at least have a worker_init (and/or container config, eventually), and we'll always want the client to have at least that default config to pass through when it submits the task

PR #362, opened 2025-10-02
Fix: issue building deployment info for function responses

hholb:
A quick fix/update to how we are building the deployment info to send down to the SDK. I wasn't tying the deployment info the the endpoint quite right, and forgot to add the conda_env_path.


OwenPriceSkelly (approved):
LGTM

PR #369, opened 2025-10-15
Add patch/delete routes for HPC entities

### LINKED ISSUE #365: Add routes for updates and deletes for HPC related data
We can create a fetch data about HPC related entites after #361. We need to add routes to update and delete HPC endpoints, deployments, and functions.

hholb:
Resolves: #365 

## Overview

This PR adds routes for updating and deleting HPC endpoints, deployments, and functions. We enforce the same rules for editing/deleting functions that we have for modal functions, and block the deletion of deployments and endpoints if there are functions associated with them.

## Testing

Lots of new tests!

I had to fight with the tests a bit. I had one test that was failing in CI, but not locally, which made it tricky to debug. After some trial and error, I think the issue was with the AWS secrets manager and the settings override. I'm not totally clear on the exact flow that cased the issue for just a single test and not any others. 🤷 

The fix was to make sure we override the `get_settings` function even when its called outside of the dependency injection system.


garden-backend-service/src/api/routes/hpc_deployments.py
```diff
@@ -22,7 +25,17 @@ async def create_hpc_deployment(
     db: AsyncSession = Depends(get_db_session),
     user: User = Depends(authed_user),
 ):
-    deployment = HpcDeployment.from_dict(deployment_data.model_dump(exclude_unset=True))
+    deployment = HpcDeployment.from_dict(
+        deployment_data.model_dump(exclude={"endpoint_ids"}, exclude_unset=True)
+    )
+
+    if deployment_data.endpoint_ids:
+        from src.models.functions.hpc.hpc_endpoints import HpcEndpoint
```
OwenPriceSkelly, line 34: nit: can we import this above with `HpcDeployment` on line 15?

garden-backend-service/src/models/functions/hpc/hpc_invocations.py
```diff
@@ -23,9 +23,13 @@ class HpcInvocationLog(Base):
...
     user_id: Mapped[int] = mapped_column(ForeignKey("users.id"))
     user: Mapped["User"] = relationship()
-    function_id: Mapped[int] = mapped_column(ForeignKey("hpc_functions.id"))
+    function_id: Mapped[int] = mapped_column(
+        ForeignKey("hpc_functions.id", ondelete="RESTRICT")
+    )
     function: Mapped["HpcFunction"] = relationship(back_populates="invocation_logs")
-    hpc_endpoint_id: Mapped[int] = mapped_column(ForeignKey("hpc_endpoints.id"))
+    hpc_endpoint_id: Mapped[int] = mapped_column(
+        ForeignKey("hpc_endpoints.id", ondelete="RESTRICT")
```
OwenPriceSkelly: understanding check: is this equivalent to how we have the relationship between modal functions and respective invocation logs set up?

OwenPriceSkelly (approved):
LGTM! (with just the one nit)

as always, love the tests. I'm pretty sure I was burned by that same kind of issue mocking the settings at some point but never connected it to the LRU cache, nice catch

garden-backend-service/src/api/routes/hpc_deployments.py
hholb: Good catch! Claude has been adding a lot of in-line imports lately, I thought I had cleaned them all up!

garden-backend-service/src/models/functions/hpc/hpc_invocations.py
hholb: We have modal functions as `SET NULL`, that is probably what we want here as well. I'll make that change since we don't want to block deletion of functions and endpoints on the condition logs exist, but do want to keep the logs around if functions and endpoints are deleted.

PR #370, opened 2025-10-20
Inlcude HPC functions in full-text search

### LINKED ISSUE #342: garden search route doesn't return matching DOIs
In some recent updates to the frontend, I changed some logic that pulls down a user's saved gardens to hit the `/gardens/search` route with list of garden DOIs. I figured this would work since the filtering logic should work for any field on a garden, but when I hit the route with a list DOIs I don't get anything back.

I am sending a request that looks like:
```json
{
  "filters": [{
    "field_name": "doi",
    "values": ["some/doi", "another/doi"],
  }]
}
```

 I think a small tweak to the filtering logic should catch the DOI filters.

Related FE issue: https://github.com/Garden-AI/garden-frontend/issues/390

### LINKED ISSUE #368: Include HPC functions in search results
Now that we have tables for HPC functions, we need to integrate them into the search route results.

hholb:
Resolves: #368 
Bonus, Resolves: #342 

## Overview

This PR adds a materialized view for text-search vectorized documents of HPC function metadata called `hpc_function_documents`. The `search_gardens` function is updated to include the HPC function documents in the search and relevance rankings. I also added helper function to refresh the materialized view, and a trigger to trigger the refresh when HPC function records are updated.

As a bonus, since it was a single line addition, I added garden DOIs in the full-text search to accommodate the use case where someone has the DOI for a garden and pastes it into the search input. Before we wouldn't return any matching gardens, now the garden with a matching DOI is the 'most relevant' result.

## Testing

Manual testing using the frontend. Gardens with HPC functions are correctly returned by the search route and all facet counts are accurate.

### DIFF
--- a/garden-backend-service/src/api/search/sql.sql
+++ b/garden-backend-service/src/api/search/sql.sql
@@ -20,16 +20,26 @@ BEGIN
          INNER JOIN modal_function_documents mf
          ON gm.modal_function_id = mf.id
          GROUP BY gm.garden_id
+    ), hpc_function_ranks AS (
+         SELECT gm.garden_id,
+                COALESCE(SUM(ts_rank(hpcf.hpcf_document, query)), 0) AS rank
+         FROM gardens_hpc_functions gm
+         INNER JOIN hpc_function_documents hpcf
+         ON gm.hpc_function_id = hpcf.id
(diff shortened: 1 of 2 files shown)


OwenPriceSkelly (approved):
LGTM!

PR #371, opened 2025-10-27
Remove hpc deployments, update other hpc schemas

### LINKED ISSUE #366: Update HPC schemas for groundhog integration

hholb:
Resolves: #369
Resolves: #366 

## Overview

This PR updates the database to remove the concept of 'hpc deployments' altogether. Now hpc function are directly associated with endpoints.

I also relaxed the deletion constraints on hpc functions and endpoints. Before we were blocking deletion if the function or endpoint was referenced in the invocation logs, now we allow deletes and just set those columns to null in the logs. This matches what we are doing for modal functions.

## Testing

Updated unit test
Manual testing via the frontend


OwenPriceSkelly (approved):
LGTM!

PR #374, opened 2025-11-04
Add owner to hpc endpoints

### LINKED ISSUE #372: Add ownership to HPC endpoints
Currently no one owns hpc endpoints. Will brought up wanting to add a custom endpoint for his own globus-compute endpoint. In that case, Will should own the endpoint and no one else should be able to edit or delete it.

Something to think about:
Do we want a visibility flag for endpoints? For example, in the above case, should all users be able to associate functions with Will's endpoint? Or does Will want to keep it 'private' so only he can associate functions with it? If it's 'private', should it still show up in metadata/search results?

hholb:
Resolves: #372 

## Overview

This PR adds and 'owner' field to hpc endpoints and updates the patch and delete routes to check that the user making the request owns the endpoint.

## Testing

New unit test!


OwenPriceSkelly (approved):
LGTM!

PR #387, opened 2025-12-09
Simplify benchmark schemas and routes

hholb:
## Overview

This PR sets up the backend for the new benchmark setup. I removed all of the job tracking and complex setup leaving a single 'BenchmarkResults' table that tracks basic information about a benchmark run with a JSON field for `metrics` meaning we can adjust the metrics we capture without changing the database schema. This setup allows us to implement more benchmarks in the SDK and only need to hit the `POST /benchmarks` to publish a run no matter what the benchmark is or what the metrics it sends contain. The frontend can hit `GET /benchmarks` to pull down data to populate the leaderboard.

The `POST` route for publising benchmark results is limited to use by super users only.

## Testing

Updated unit tests.


WillEngler (approved):
Love to see a net red PR!

